PR #7, opened 2019-10-14
Create constants for slots and slot groups

ChloeDawn:
Also adds `TrinketComponent#getStack(String, String)` which is implemented as
```java
public default ItemStack getStack(String group, String slot){
	return getStack(group + ':' + slot);
}
```

### DIFF
--- a/src/main/java/dev/emi/trinkets/api/PlayerTrinketComponent.java
+++ b/src/main/java/dev/emi/trinkets/api/PlayerTrinketComponent.java
@@ -69,7 +69,7 @@ public class PlayerTrinketComponent implements TrinketComponent, EntitySyncedCom
 			if(stack.getItem() == Items.ELYTRA){
 				for(SlotGroup group: TrinketSlots.slotGroups){
 					for(Slot slot: group.slots){
-						if(group.getName().equals("chest") && slot.getName() == "cape"){
+						if(group.getName().equals(SlotGroups.CHEST) && slot.getName() == "cape"){
 							if(getInventory().getInvStack(i).isEmpty()){
 								getInventory().setInvStack(i, stack.copy());
 								TrinketsClient.lastEquipped = group;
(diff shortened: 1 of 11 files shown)


TehNut:
This would be a breaking change

ChloeDawn:
@TehNut Resolved

src/main/java/dev/emi/trinkets/mixin/ClientPlayerEntityMixin.java
```diff
@@ -32,6 +34,6 @@ public void closeScreen(CallbackInfo info){
 	@Redirect(at = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerEntity;getEquippedStack(Lnet/minecraft/entity/EquipmentSlot;)Lnet/minecraft/item/ItemStack;"), method = "tickMovement")
 	public ItemStack getEquippedStackProxy(ClientPlayerEntity player, EquipmentSlot slot){
 		TrinketComponent comp = TrinketsApi.getTrinketComponent(player);
-		return comp.getStack("chest:cape");
+		return comp.getStack(SlotGroups.CHEST + ':' + Slots.CAPE);
```
emilyploszaj: This method call is still concatenating strings rather than using the function that was added like the others.

src/main/java/dev/emi/trinkets/api/PlayerTrinketComponent.java
```diff
@@ -69,7 +69,7 @@ public boolean equip(ItemStack stack){
 			if(stack.getItem() == Items.ELYTRA){
 				for(SlotGroup group: TrinketSlots.slotGroups){
 					for(Slot slot: group.slots){
-						if(group.getName().equals("chest") && slot.getName() == "cape"){
+						if(group.getName().equals(SlotGroups.CHEST) && slot.getName() == Slots.CAPE){
```
emilyploszaj: This has inconsistent comparison still.

ChloeDawn: I considered this out of scope for my changes, I do not fix string comparisons elsewhere.

src/main/java/dev/emi/trinkets/mixin/ClientPlayerEntityMixin.java
ChloeDawn: Resolved in 3092963

src/main/java/dev/emi/trinkets/api/PlayerTrinketComponent.java
emilyploszaj: valid

PR #20, opened 2020-05-01
Add methods ITrinket#onEquipServer and ITrinket#onUnequipServer

bluepython508:
ITrinket#onEquip and ITrinket#onUnequip are only called on the client side, which makes making items that modify the player when they are equipped much more difficult. This fix shouldn't break backwards compatibility, as ITrinket#onEquipServer and ITrinket#onUnequipServer are both defaulted methods, with an empty default implementation, like ITrinket#onEquip and ITrinket#onUnequip. The documentation comments for ITrinket#onEquip and ITrinket#onUnequip have also been updated to make this divide clearer.

### DIFF
--- a/src/main/java/dev/emi/trinkets/TrinketsMain.java
+++ b/src/main/java/dev/emi/trinkets/TrinketsMain.java
@@ -6,16 +6,18 @@ import dev.emi.trinkets.api.SlotGroups;
 import dev.emi.trinkets.api.Slots;
 import dev.emi.trinkets.api.TrinketSlots;
 import dev.emi.trinkets.api.TrinketsApi;
+import dev.emi.trinkets.api.TrinketInventory;
 import nerdhub.cardinal.components.api.event.EntityComponentCallback;
 import nerdhub.cardinal.components.api.util.EntityComponents;
 import nerdhub.cardinal.components.api.util.RespawnCopyStrategy;
 import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.network.ServerSidePacketRegistry;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.item.Items;
 import net.minecraft.util.Identifier;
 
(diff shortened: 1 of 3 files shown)


emilyploszaj:
onEquip and onUnequip *should* be working server-side already, or at least they did at some point. I don't think an extraneous packet is needed or desirable for this fix since it should be easy to just  add a one line fix in https://github.com/emilyploszaj/trinkets/blob/master/src/main/java/dev/emi/trinkets/mixin/PlayerEntityMixin.java (where equipping and unequipping is already handled for attributes). If you want to rework your pull to just commit that one line that's cool but otherwise I'll just add it and commit it myself. Thanks for pointing this out though!

bluepython508:
When I tested, they were only called on the client, and when I gave them @Environment(EnvType.SERVER) they didn’t get called at all. Are you suggesting a call to onEquip and onUnequip in PlayerEntityMixin addAttributes and removeAttributes?

emilyploszaj:
Well, not necessarily in those functions but the same place where they get called, since that already keeps track of when trinket slots change

bluepython508:
That makes sense. Would it be better to have separate client and server methods, or just one called on both sides?

bluepython508:
Alright, this makes no sense anymore. Apparently, when onEquip or onUnequip is @Environment(EnvType.Server), it doesn't get called at all, but when it doesn't have an environment annotation, it is called on both sides. Also, both methods are called several times on the client each time the item is removed or added. At least where I added onEquipServer and onUnequipServer, they're only called on the server, and only once per equip or unequip.

emilyploszaj:
Ah, yeah I seem to remember that. I'm pretty sure it has something to do with the client doing a gui action, telling the server about it, and the server telling the client that the action took place, it's unfortunate *but* the client always ends up with a basically equivalent set of events (ie EUE instead of just E) so I think it's "fine". Since this is the case and it seems to be working, I'm going to close this PR.

PR #30, opened 2020-07-18
Fix for dispensers chewing up items

Avekeez:
Previously, the included dispenser behaviour for TrinketItem would chew up the stack upon failure to find a valid entity to pick up the trinket, as well as completely empty a stack when dispensing a stackable trinket - this was fixed to dispense only one item from a stack a time and fire it from the dispenser upon failure to find a target. In addition, this fixes trinket slots not being considered when trying to find a spot to equip, and so failing when the inventory is full even if the trinket slot is open - this fix kicks in for any item implementing Trinket, so I also made ElytraItem do so to avoid having to make an exception for it.


emilyploszaj:
wow I can't believe the initial behavior was this broken, thanks for the fix, I'll get this in a release soon.

PR #35, opened 2020-08-07
Fix crash when other mods add to `HandledScreen.handler.slots`

camas:
Stores the offset of the first TrinketSlot and uses that to calculate future offsets instead of using a hardcoded value

Assumes slots are stored sequentially

Fixes https://github.com/camas/inv-trash-slot/issues/2


emilyploszaj:
There may be a more elegant way for me to fix this in the future but for now this will work, thanks

PR #37, opened 2020-09-05
Basic data-driven slots

TheIllusiveC4:
- Created `SlotLoader` and `EntitySlotLoader` to load slots and entity slot distribution respectively
- Registered data listeners in `TrinketsMain`
- Created `TrinketEnums` to hold various enums, currently has `DropRule` for denoting drop behavior to be fully implemented at a later time
- Created `SlotGroup` and `SlotType` for slot groups and slots respectively
- Created `TrinketSlots` as the primary way to access entity slots
- Added data files for Trinkets's list of suggested slots
- Added myself to the authors list in `fabric.mod.json`
- Added entrypoints for CCA static component registration (needed this in order to fully test this PR)

The data loading is still rudimentary, notably lacking proper error handling, but it should provide enough functionality to move forward for the moment. As long as slots are accessed primarily through `TrinketSlots`, the implementation details can change later on without affecting the rest of the code.

One benefit of this system is that we can define suggested slots and they will exist, but they won't do anything until developers or users define an entity to receive those slots. I have defined all of the slots that Trinkets has historically suggested and given them their default icons.

Here's a datapack that we can use to add all of those slots to the player, for testing purposes, so that we never have to attach any slots to the player natively.

[debug.zip](https://github.com/emilyploszaj/trinkets/files/5178615/debug.zip)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ bin/
 # fabric
 
 run/
+logs/
--- a/src/main/java/dev/emi/trinkets/TrinketsClient.java
+++ b/src/main/java/dev/emi/trinkets/TrinketsClient.java
@@ -5,7 +5,7 @@ import net.fabricmc.api.ClientModInitializer;
 public class TrinketsClient implements ClientModInitializer {
 
 	@Override
-	public void onInitializeClient(){
+	public void onInitializeClient() {
 
 	}
 }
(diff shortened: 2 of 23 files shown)


emilyploszaj (changes requested):
Looks good content wise but most changes here use two spaces instead of tabs which don't line up with existing code, you could fix that, I could merge a style fix after this, or we could live with dissonant spacing forever, whatever you'd prefer

emilyploszaj (approved):

PR #40, opened 2020-09-23
Streamline data handling

TheIllusiveC4:
- Add error handling and logging for data reading
- Remove `TrinketSlots` and move `getPlayerSlots` and `getEntitySlots` to `TrinketsApi`
  - I couldn't think of enough reasons to keep `TrinketSlots` around, so I just removed it to keep the API package from being too cluttered. But I can also put it back if we want, I don't really have a strong opinion on that.
- `getEntitySlots` now works properly with the player entity type


emilyploszaj (approved):
I think it does make sense to move it all into `TrinketsApi` but if we want to move it around later we definitely have time

PR #41, opened 2020-09-28
Small refactors and trinket renderer registry

TheIllusiveC4:
* Move the Trinket `ComponentKey` from `TrinketsMain` to `TrinketsApi`
* Refactor `getTrinket` to return an `Optional` value to designate to developers that they need to be aware of obtaining empty states
* Add `getTrinketComponent` that basically wraps CCA's `Component.maybeGet` function for the trinket component and refactor some existing logic to use this new function
* Set up the logic for registering trinket renderers and rendering them, mainly in `TrinketRenderer`, `TrinketRendererRegistry`, and `PlayerEntityRendererMixin`

### DIFF
--- a/src/main/java/dev/emi/trinkets/api/client/TrinketRenderer.java
+++ b/src/main/java/dev/emi/trinkets/api/client/TrinketRenderer.java
@@ -13,6 +13,7 @@ public interface TrinketRenderer {
 	void render(ItemStack stack, Trinket.SlotReference slot, MatrixStack matrixStack, VertexConsumerProvider vertexConsumer, int light,
 			PlayerEntityModel<AbstractClientPlayerEntity> model, AbstractClientPlayerEntity player, float headYaw, float headPitch);
 
+	// todo: Review these translation methods before release
 	/**
 	 * Translates the rendering context to the center of the player's face
 	 */
--- a/src/main/java/dev/emi/trinkets/api/client/TrinketRenderer.java
+++ b/src/main/java/dev/emi/trinkets/api/client/TrinketRenderer.java
@@ -1,0 +1,111 @@
+package dev.emi.trinkets.api.client;
+
+import dev.emi.trinkets.api.Trinket;
+import net.minecraft.client.network.AbstractClientPlayerEntity;
+import net.minecraft.client.render.VertexConsumerProvider;
+import net.minecraft.client.render.entity.model.PlayerEntityModel;
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.client.util.math.Vector3f;
+import net.minecraft.item.ItemStack;
+
+public interface TrinketRenderer {
+
+	void render(ItemStack stack, Trinket.SlotReference slot, MatrixStack matrixStack, VertexConsumerProvider vertexConsumer, int light,
+			PlayerEntityModel<AbstractClientPlayerEntity> model, AbstractClientPlayerEntity player, float headYaw, float headPitch);
+
+	/**
+	 * Translates the rendering context to the center of the player's face
+	 */
+	static void translateToFace(MatrixStack matrixStack, PlayerEntityModel<AbstractClientPlayerEntity> model, AbstractClientPlayerEntity player, float headYaw,
+			float headPitch) {
+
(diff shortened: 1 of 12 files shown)


src/main/java/dev/emi/trinkets/api/client/TrinketRendererRegistry.java
```diff
@@ -0,0 +1,19 @@
+package dev.emi.trinkets.api.client;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import net.minecraft.item.Item;
+
+public class TrinketRendererRegistry {
```
emilyploszaj, line 8: Is there a reason to have this as a separate file from `TrinketsApi` other than separating server/client? I'm fine with this but I just want to hear your thought process.

src/main/java/dev/emi/trinkets/mixin/PlayerEntityRendererMixin.java
```diff
@@ -0,0 +1,28 @@
...
+import net.minecraft.client.network.AbstractClientPlayerEntity;
+import net.minecraft.client.render.entity.EntityRenderDispatcher;
+import net.minecraft.client.render.entity.LivingEntityRenderer;
+import net.minecraft.client.render.entity.PlayerEntityRenderer;
+import net.minecraft.client.render.entity.model.PlayerEntityModel;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
```
emilyploszaj, line 15: I'd prefer a comment here explaining the mixin, even if it is quite simple

src/main/java/dev/emi/trinkets/api/client/TrinketRenderer.java
```diff
@@ -0,0 +1,111 @@
...
+
+import dev.emi.trinkets.api.Trinket;
+import net.minecraft.client.network.AbstractClientPlayerEntity;
+import net.minecraft.client.render.VertexConsumerProvider;
+import net.minecraft.client.render.entity.model.PlayerEntityModel;
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.client.util.math.Vector3f;
+import net.minecraft.item.ItemStack;
+
+public interface TrinketRenderer {
```
emilyploszaj, line 11: We might want to review these translations at some point, perhaps a translate to main arm and secondary arm rather than right and left? Or perhaps both?

emilyploszaj (changes requested):
Just a couple small things

src/main/java/dev/emi/trinkets/api/client/TrinketRendererRegistry.java
TheIllusiveC4: It's mainly just to separate server/client logic. More particularly, just making sure the client-only classes remain separate from everything else to avoid class-loading issues on servers. Originally I even had this named as `TrinketsClientApi` for putting any client-only logic, but I wasn't really sure if we were going to have anything that falls under this category other than this rendering code.

src/main/java/dev/emi/trinkets/api/client/TrinketRenderer.java
TheIllusiveC4: Hmm, fair point. Since I'd imagine this might be lower priority, should I just mark it with a to-do for now?

src/main/java/dev/emi/trinkets/api/client/TrinketRendererRegistry.java
emilyploszaj: We can rename it if we need to sounds good for now.

src/main/java/dev/emi/trinkets/api/client/TrinketRenderer.java
emilyploszaj: Sounds good to me, we can discuss in dm at some later point before release.

emilyploszaj (approved):

PR #42, opened 2020-10-21
Update to Cardinal Components API V3

Pyrofab:
With CCA V2 interfaces getting removed in 1.17, now is a fairly good time to switch to V3. The changes are not too impressive, the main one being the [registration mechanism](https://github.com/OnyxStudios/Cardinal-Components-API/wiki/Registering-and-using-a-component).

`ComponentType` is supposed to be replaced with `ComponentKey`, but that would break binary compatibility with the mods that use this API; so I left it with a bit of an awkward cast. You can probably fix that kludge in your own 3.0.0 update.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,11 +35,11 @@ dependencies {
 	// You may need to force-disable transitiveness on them.
 	
 	//Base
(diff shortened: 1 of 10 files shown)


Pyrofab:
OK turns out the CCA update is already part of the 3.0.0 branch, should have checked more thoroughly

PR #48, opened 2020-11-24
Add slot ID data for vanilla slot groups

TheIllusiveC4:
Pretty straightforward as we discussed in DMs:

* Adds a data field for group json files to specify a slot ID used as the default slot. This is relevant mainly as an internal way for us to access the slot ID to overlay for the vanilla inventory slot groups we provide, but it's probably possible for external use in custom GUI implementations as well.
* Adds a group json file for each of the slots we provide with the proper group information. Head, chest, legs, feet, and offhand define its default slot with a slot ID matching the inventory slot (Note: I got these slot IDs from the master branch. I can't imagine that they've changed, but I didn't test their validity). Hands defines its default slot with a string-based slot name.
* Slot ID fields are -1 if not defined.


emilyalexandra (approved):
As you said, pretty straightforward, we both opened pulls at the same time so I'm gonna have to resolve the merge conflict this is going to cause when I merge this on my pr

PR #52, opened 2020-12-03
Trinket EAMs and shift clicking

emilyalexandra:
Updated to 1.16.4

Trinket EAMs are now handled
* The signature for getting modifiers now passes a UUID to use

Shift clicking now goes into trinket slots
* Uses the quick move predicate system
* Does the classic trinket slot group preview when a slot is shift clicked into (showing the group for 20 ticks unless another group is focused in that time)

Smaller things
* Trinket `onEquip` and `onUnequip` methods are now called, meaning every method in `Trinket` is now handled and used properly (except for `onBreak`)
* Buttons are no longer highlighted or interactable while a trinket group is open (a bug in the classic trinkets that was never actually fixed)

One final note, we're getting really close to release ready with trinkets, I would consider it very usable at this state. The biggest things that are unimplemented are locked slots and the fact that slot amounts are handled as different slots instead of expanding out vertically like we mentioned.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,12 +2,12 @@ name=trinkets
 # Done to increase the memory available to gradle.
 org.gradle.jvmargs=-Xmx1G
 
-minecraft_version=1.16.2
-yarn_mappings=1.16.2+build.19
-loader_version=0.9.1+build.205
+minecraft_version=1.16.4
+yarn_mappings=1.16.4+build.7
+loader_version=0.10.8
 
 mod_version = 3.0.0
 maven_group = dev.emi
 archives_base_name = trinkets
 
-fabric_version=0.18.0+build.397-1.16
+fabric_version=0.27.1+1.16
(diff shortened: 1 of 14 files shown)


emilyalexandra:
I've also just added a couple things I forgot to do, that being a refactor of the trinket render system and some helper method for getting or determining equipped items

gradle.properties
```diff
@@ -2,12 +2,12 @@ name=trinkets
 # Done to increase the memory available to gradle.
 org.gradle.jvmargs=-Xmx1G
 
-minecraft_version=1.16.2
-yarn_mappings=1.16.2+build.19
-loader_version=0.9.1+build.205
+minecraft_version=1.16.4
```
TheIllusiveC4, line 5: Is there a reason we're building against 1.16.4 instead of one of the snapshot versions since we're targeting 1.17? Not that I think it really matters for our purposes, but I'm just curious.

TheIllusiveC4 (approved):

gradle.properties
emilyalexandra: A number of people have said they'd want to use 3.0.0 on 1.16.4, so I was planning to do the upgrade last if at all, I'm thinking of potentially changing the plan we originally had and converting all mods that use trinkets to 3.0.0 manually and pr that stuff in, since 7 months is a long wait

PR #54, opened 2020-12-08
Slot syncing

TheIllusiveC4:
Implements slot syncing. Also fixes the bug with `/reload`.
- Adds `TrinketsNetwork` class for storing packet identifiers
- Registers the new sync packet to the client packet registry
- Adds serialization and deserialization helper methods to `SlotGroup` and `SlotType`
- Adds syncing helper methods to `EntitySlotLoader`
- Adds `PlayerManagerMixin` to hook syncs to players logging in (I don't think Fabric API has a callback for this?)
- Hooks syncing logic to the `END_DATA_PACK_RELOAD` event to fix the `/reload` issue
- Extracts and edits some of the updating logic in `TrinketInventory`, which was the cause of the slots disappearing behavior because it was still holding old references once the slots updated

I've tested with `/reload`, logging in and out, and on dedicated servers, and everything seems to check out, but I haven't tested anything beyond that.

### DIFF
--- a/src/main/java/dev/emi/trinkets/TrinketPlayerScreenHandler.java
+++ b/src/main/java/dev/emi/trinkets/TrinketPlayerScreenHandler.java
@@ -9,7 +9,7 @@ import net.minecraft.util.Pair;
 public interface TrinketPlayerScreenHandler {
 	
 	/**
-	 * Called to inform the player's slot handler that it needs to remove and read its trinket slots to reflect new changes
+	 * Called to inform the player's slot handler that it needs to remove and re-add its trinket slots to reflect new changes
 	 */
 	public void updateTrinketSlots();
 
--- a/src/main/java/dev/emi/trinkets/TrinketPlayerScreenHandler.java
+++ b/src/main/java/dev/emi/trinkets/TrinketPlayerScreenHandler.java
@@ -9,7 +9,7 @@ import net.minecraft.util.Pair;
 public interface TrinketPlayerScreenHandler {
 	
 	/**
-	 * Called to inform the player's slot handler that it needs to remove and readd its trinket slots to reflect new changes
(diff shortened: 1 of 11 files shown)


src/main/java/dev/emi/trinkets/TrinketPlayerScreenHandler.java
```diff
@@ -9,7 +9,7 @@
 public interface TrinketPlayerScreenHandler {
 	
 	/**
-	 * Called to inform the player's slot handler that it needs to remove and readd its trinket slots to reflect new changes
+	 * Called to inform the player's slot handler that it needs to remove and read its trinket slots to reflect new changes
```
emilyalexandra, line 12: This was supposed to be "re-add" as in "add again"

src/main/java/dev/emi/trinkets/mixin/PlayerManagerMixin.java
```diff
@@ -0,0 +1,25 @@
...
+ * Syncs slot data to player's client on login
+ *
+ * @author C4
+ */
+@Mixin(PlayerManager.class)
+public class PlayerManagerMixin {
+
+	@Inject(at = @At("TAIL"), method = "onPlayerConnect")
+	public void onPlayerConnect(ClientConnection connection, ServerPlayerEntity player, CallbackInfo cb) {
+      EntitySlotLoader.INSTANCE.sync(player);
```
emilyalexandra, line 23: This is indented improperly, and the github is complaining about the lack of a trailing newline on this file

emilyalexandra (commented):
Just some minor fixes I'm going to add myself, one thing I noticed while reading through is that all non-player living entities will run into some desync because their inventories are not updated (though that can be fixed by running the `TrinketInventory.update()` provided on all living entities on reload, so I'll look into that later

PR #58, opened 2021-01-26
[3.0.0] Refactor/cleanup, test mod and dependency updates

florensie:
Refactors include:
- cleaning up imports
- removing unnecessary field/method modifiers in interfaces
- updating the build.gradle so it's in line with the current fabric example mod
- Making several fields final where possible in some places
- Updating deprecated api usage (cca and fabric-api networking)
- Updated fabric.mod.json

Additions:
- `getAllEquipped` shortcut method on `TrinketComponent`
- test mod to easily test in-game (set classpath in IDE run config to `trinkets.testmod`)


src/main/java/dev/emi/trinkets/TrinketsNetwork.java
```diff
@@ -4,6 +4,8 @@
 
 public class TrinketsNetwork {
 
-  public static final Identifier SYNC_SLOTS = new Identifier(TrinketsMain.MOD_ID, "sync_slots");
+  private TrinketsNetwork() {
```
emilyalexandra, line 7: I tend to prefer fields above methods, including constructors, is this a different idiom for private constructors?

src/main/java/dev/emi/trinkets/api/LivingEntityTrinketComponent.java
```diff
@@ -98,12 +98,12 @@ public boolean isEquipped(Predicate<ItemStack> predicate) {
 
 	@Override
 	public List<Pair<SlotReference, ItemStack>> getEquipped(Predicate<ItemStack> predicate) {
-		List<Pair<SlotReference, ItemStack>> list = new ArrayList<Pair<SlotReference, ItemStack>>();
```
emilyalexandra, line 101: What's the purpose of erasing generics here?

src/main/java/dev/emi/trinkets/api/TrinketComponent.java
```diff
@@ -1,27 +1,31 @@
...
-	public List<Pair<SlotReference, ItemStack>> getEquipped(Predicate<ItemStack> predicate);
+	List<Pair<SlotReference, ItemStack>> getEquipped(Predicate<ItemStack> predicate);
 
-	public default List<Pair<SlotReference, ItemStack>> getEquipped(Item item) {
+	default List<Pair<SlotReference, ItemStack>> getEquipped(Item item) {
 		return getEquipped(stack -> stack.getItem() == item);
 	}
+
+	default List<Pair<SlotReference, ItemStack>> getAllEquipped() {
+		return getEquipped(stack -> true);
```
emilyalexandra, line 29: This is going to return all slots, even empty ones, is this intended behavior or would `stack -> !stack.isEmpty()` be a better option?

src/main/java/dev/emi/trinkets/api/TrinketsApi.java
```diff
@@ -21,16 +15,20 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
 public class TrinketsApi {
 
 	public static final ComponentKey<TrinketComponent> TRINKET_COMPONENT = ComponentRegistryV3.INSTANCE
 			.getOrCreate(new Identifier(TrinketsMain.MOD_ID, "trinkets"), TrinketComponent.class);
 	private static final Map<Identifier, Function3<ItemStack, SlotReference, LivingEntity, TriState>> QUICK_MOVE_PREDICATES
-			= new HashMap<Identifier, Function3<ItemStack, SlotReference, LivingEntity, TriState>>();
```
emilyalexandra, line 29: Same question here as before, what is erasing generics doing other than reducing line size?

src/main/java/dev/emi/trinkets/api/client/TrinketRenderer.java
```diff
@@ -9,7 +9,7 @@
 import net.minecraft.util.math.Vec3f;
 
 public interface TrinketRenderer {
-	public static final float MAGIC_ROTATION = 180f / (float) Math.PI;
+	float MAGIC_ROTATION = 180f / (float) Math.PI;
```
emilyalexandra, line 12: It feels weird to not have this annotated with `final`, especially as a capitalized field.

emilyalexandra (changes requested):
Thanks for figuring out the decimal format thing, lol, I probably gave up on that too quickly. Overall pretty good but I'm curious about some widespread changes, like erasing generics in places (is this a style preference I was not aware of?). Also thanks for teaching me that `@Unique` exists, I've probably polluted fields all over the place in my mixins. Also, I know it was a relatively straightforward change, but just to make sure, you tested the new packet system on a dedicated server with different trinket slots from the client, right?

src/main/java/dev/emi/trinkets/api/client/TrinketRenderer.java
florensie: All members (both fields and methods) of an interface are implicitly public, which explains why I removed `public` from all the interface methods.

Additionally, all fields of an interface are always static and final. See [this stackoverflow post](https://stackoverflow.com/questions/1513520/why-are-all-fields-in-an-interface-implicitly-static-and-final) for a good explanation.

src/main/java/dev/emi/trinkets/api/TrinketComponent.java
florensie: Good catch

src/main/java/dev/emi/trinkets/api/client/TrinketRenderer.java
emilyalexandra: I understood that this was implicit from the change but I just meant it looks weird to have allcaps without final, probably because I'm not used to it

src/main/java/dev/emi/trinkets/TrinketsNetwork.java
florensie: I generated it with my IDE and for some reason it put it there by default.

src/main/java/dev/emi/trinkets/api/LivingEntityTrinketComponent.java
florensie: It's a well known feature in java that in an assignment like these you can leave them out. Having it twice is redundant and doesn't read any better imo. Tbh I was just fixing IDE warnings here.

emilyalexandra: Sounds good to me, since you changed it in the whole codebase I don't have to worry about inconsistency and will just switch to writing like this, it saves me time

src/main/java/dev/emi/trinkets/api/client/TrinketRenderer.java
florensie: Ah I can see why. From looking around a bit though this seems pretty standard.

florensie:
I haven't gotten to any real testing yet as I'm not entirely sure what's currently supposed to be working already in this rewrite. Maybe it would be a good idea to add a features list in the readme with checkboxes indicating what's complete? In the test mod I added a trinket but it doesn't seem like any of the functionality is working from a user's standpoint (equipping, gui, tooltips).

emilyalexandra:
Nearly everything is in working order, slots are data driven, they sync, the gui works as expected, stuff like trinket attributes are functioning. The main things that aren't working are breaking particles (not sure exactly how they're to work, it was C4's idea), xp repairing mending trinkets (just realized this while typing), and treating duplicate slots (amount in slot type) so they extend vertically instead of add to the list horizontally (as a sort of qol space saving thing). Basically everything else is in working order.

emilyalexandra (approved):

florensie:
I'll see if I can write something in the test mod that actually works tomorrow.

PR #60, opened 2021-01-27
[3.0.0] Changes to default slots

florensie:
This PR adds two slots. I originally made these for my own mod but I think these make a lot of sense for a lot of other mods too.
- Hat: makes a lot of sense since there have been a bunch of hat mods that use trinkets
- Shoes: in my opinion `aglets` is too specific and shoes make more sense. I've kept `aglets` in for now but I'm personally in favor of removing it as it seems like it's usage would be specific to a single mod. If agreed I can remove `aglets` in this PR.

This PR also changes the gloves slot:
- I have split up `gloves` into a slot (`glove`) for both `hand` and `offhand`, just like `ring`
- Tweaked the glove slot icon texture, feel free to leave your thoughts about it


emilyalexandra:
I think both hat and shoes are fine slot types to add (though I think hat is more useful than shoes). They both pose a weird logistical issue imagining how one could wear a helmet with a hat or shoes with boots... but creative interpretation is required for capes and backpacks as well. These seem fine to me, I'll see what C4 thinks.

On to proposed changes of existing slots. For gloves, I think it makes more sense to have gloves be a single slot. Most of the time when gloves are implemented, they're a pair of gloves in a single item, and making mod makers make 2 gloves and balance around this seems weird, I'd like to see C4's take. I do understand how it makes more sense conceptually to have a gloves mirror rings for the hand/offhand parity

For aglets, this is mostly a personal thing, but I like the character they bring to the API. It's nothing huge but I liked the push to make me design aglets (instead of something more simple) when I was working on bunch o' trinkets. Of course, like other slots, this doesn't have a wide usage but I think there is no harm in recommending a mildly silly slot that adds a bit of character to hopefully both the API and the fabric ecosystem in genera.

As I said before, I'd also like to see what C4 thinks about these proposed changes before I make a decision, if you have any comments on what I've said, feel free.

florensie:
> Of course, like other slots, this doesn't have a wide usage but I think there is no harm in recommending a mildly silly slot that adds a bit of character to hopefully both the API and the fabric ecosystem in general.

Oh if it's like that haha. I guess I didn't see the humor in it, but then again I also didn't know what an aglet was until I looked it up. I guess it can be Trinket's tiny potato. :smile: Like you say it definitely doesn't do any harm since it needs to be enabled anyways.

As for gloves, I guess I'm biased because that's how gloves currently work in my own mod:
```java
TrinketSlots.addSlot(SlotGroups.HAND, Slots.GLOVES,
		new Identifier("trinkets", "textures/item/empty_trinket_slot_gloves.png"));
TrinketSlots.addSlot(SlotGroups.OFFHAND, Slots.GLOVES,
		new Identifier("trinkets", "textures/item/empty_trinket_slot_gloves.png"));
```
Part of the problem is that when you _do_ want separate gloves in your mod it gets kind of confusing:
- There is already a slot called `hand:gloves` but it's only on the main hand so do you add `offhand:gloves` or do you add two new slots for each hand called `glove` (singular)?
- If there's a mod that uses combined gloves and the player puts the combined one in `hand:gloves` they can still put a singular glove into my mod's `offhand:gloves` slot and the rendering will overlap

It would almost make sense to create a group called `hands` that's used for both and has some sort of subgroup for each hand but I'm not sure how that would be implemented and displayed correctly or if it's even worth the time figuring that out.

TheIllusiveC4:
I think shoes and hat slots are both fine to add, I can imagine quite a few mods implementing such items.

I also agree to keep the aglet slot. It might be rather specific, but there's no harm in keeping it and Emi brings up a good point in that it could inspire new trinket designs in the ecosystem.

The topic of `gloves` to `glove` is a bit trickier. I don't think rendering overlaps should be a major concern since it will always occur one way or another due to the dynamic nature of the system. This could only be avoided if we either force a single slot or force developers to separate gloves, neither of which are practical.

One way to look at this issue is to imagine a situation where Mod A wants a consolidated glove slot and Mod B wants separate glove slots and they're both in the same environment. No matter which standard we choose, the end result will be that there's two slots added because Mod B needs them. Whether Mod A has to choose between one slot or two slots is irrelevant because the second slot is only needed when Mod B is present.

Given that situation, I think I lean towards agreeing with the change to separate them. Mods that just need a consolidated `gloves` slot can just use `hand:glove`, and mods that need both will use both. It also seems a bit more intuitive because of how rings are already setup like this.

emilyalexandra:
I think I'll go with C4's judgement here.

Now that we've decided what slots to add, I do want to ask about icons. I prefer my original glove icon to the one you've made (it looks a lot like an oven mitt to me) but I'd also be fine with having a go with designing a new glove icon all together. I am fine with the hat and shoes icons. Though, I may try around with adding more of a defined rim to the hat.

TheIllusiveC4:
Hmm, I'm not sure which icon I prefer. They're probably about equal in my eyes. I do agree that the proposed one looks a tad like an oven mitt, but also the original one kind of looks more like a `claw` or `hand` to me than a `glove` slot. Admittedly it's hard to design something that specifically looks like a glove without walking into oven mitt territory. I'd be fine with either, but I'm also open to the idea of a new icon entirely.

florensie:
When I was trying a couple of things I also made this one. It's more of a traditional hand shape with distinct fingers. It's still not perfect but it might be more the direction it should go.
glove1

florensie:
Here's another iteration for a glove icon. This time I just took the glove icon [I already had](https://github.com/ochotonida/artifacts/blob/a97b1a92fd249772cb62d6355f334112966bf0cc/src/main/resources/assets/artifacts/textures/item/fire_gauntlet.png), outlined it and tweaked it a bit.
glove3

emilyalexandra:
I like the first more than the second, but I'm wondering if you can give it more of an angle? it feels very square when oriented like that

florensie:
glove3
This is the first one with some elements of the second to make it a bit less square.

emilyalexandra:
I think I prefer the first one, and it'd be fine as a final image, I tried drawing some stuff up but couldn't get anything I liked more than that, if you've got any more I'd be willing to check them out

florensie:
I'll just stick with that one then for now. Feel free to try your _hand_ at it of course.

emilyalexandra:
Here's one last idea, what if the offhand slot used a horizontally mirrored version of the glove texture, since they are supposed to represent different hands?

TheIllusiveC4:
I'd be okay with that.

PR #61, opened 2021-01-31
Javadocs, rendering & testmod item

florensie:
- Update and add javadocs
- Rendering with LivingEntities and EntityModels
- followBodyRotations [from Curios](https://github.com/TheIllusiveC4/Curios/blob/d107f06b020a4f94e9e6dad374daf74d1f8e0de3/src/main/java/top/theillusivec4/curios/api/type/component/IRenderableCurio.java#L149) but slightly modified to only take one `BipedEntityModel`
This is basically an all in one for the `translateTo*` methods, should those be kept?
- Fix some possible bugs
- Add working trinket to test mod


src/testmod/java/dev/emi/trinkets/TestTrinket.java
```diff
@@ -1,10 +1,27 @@
...
 public class TestTrinket extends TrinketItem {
 
+	private static final EntityAttributeModifier SPEED_BOOST_MODIFIER = new EntityAttributeModifier(UUID.fromString("ac7ab816-2b08-46b6-879d-e5dea34ff305"), "trinkets-testmod:movement_speed", 0.4, EntityAttributeModifier.Operation.MULTIPLY_TOTAL);
+
 	public TestTrinket(Settings settings) {
 		super(settings);
 	}
+
+	@Override
+	public Multimap<EntityAttribute, EntityAttributeModifier> getModifiers(ItemStack stack, SlotReference slot, LivingEntity entity, UUID uuid) {
```
emilyalexandra, line 22: This UUID behavior may be a holdover from experience you may have in curios, but that's actually a behavior trinkets corrects, you should be constructing an `EntityAttributeModifier` with the passed UUID.

Explanation if you want to know:
Using a constant UUID means that your effect won't stack, and only apply once, worse, if you wear 2 of the same trinket and then remove one, you will not get the effect at all (this is because unequiping removes the EAM and minecraft tracks EAMs using UUIDs). There are also some other edge cases that will result in unintended behavior. This is the sort of thing you don't have to worry about with armor slots or held items because only a single instance can be applied at once.

emilyalexandra (changes requested):
Other than the EAM UUID issue, this is good. Particularly, me and C4 were talking about fixing the renderer being player entity specific. I think it's useful to keep the translate methods, as different approaches without models (like rendering items or blocks) would probably appreciate it.

src/main/java/dev/emi/trinkets/api/client/TrinketRenderer.java
```diff
@@ -1,19 +1,60 @@
...
+	/**
+	 * Rotates the rendering for the models based on the entity's poses and movements. This will do
+	 * nothing if the entity render object does not implement {@link LivingEntityRenderer} or if the
+	 * model does not implement {@link BipedEntityModel}).
+	 *
+	 * @param entity The wearer of the trinket
+	 * @param model The model to align to the body movement
+	 */
+	static void followBodyRotations(final LivingEntity entity,
+									final BipedEntityModel<LivingEntity> model) {
```
TheIllusiveC4, line 41: Is there a particular reason that this was modified from Curios to take only one model?

src/testmod/java/dev/emi/trinkets/TestTrinket.java
florensie: Oh, that makes a lot of sense, thanks for the explanation.

src/main/java/dev/emi/trinkets/api/client/TrinketRenderer.java
florensie: Correct me if I'm wrong but when you create a BipedModel this way for your trinket I don't expect anyone to supply more than one model, ever. I just don't really see any use for it and maybe makes the api a bit less confusing and it's implementation a bit more readable.

florensie:
> I think it's useful to keep the translate methods, as different approaches without models (like rendering items or blocks) would probably appreciate it.

@emilyalexandra that's the only reason I kept them in for now, though I do still think they might need some work (if possible) as they'll only properly work with player models because of those hardcoded values.

src/testmod/java/dev/emi/trinkets/TestTrinket.java
florensie: > This UUID behavior may be a holdover from experience you may have in curios

Also correct on this btw 😉

src/main/java/dev/emi/trinkets/api/client/TrinketRenderer.java
TheIllusiveC4: That's fair. I was just mainly curious if there was a more major reason for it, I think that's fine as it is. It's just a helper method anyway, so it doesn't necessarily need to cover every use-case.

src/testmod/java/dev/emi/trinkets/TestTrinket.java
TheIllusiveC4: I think this is less a holdover from any Curios experience (although it certainly doesn't help) and more to do with how modders are accustomed to setting up their EAMs in static initializers since vanilla doesn't need this sort of consideration. While I worry this will happen fairly often, I'm not sure how to make this more clear to modders since the javadocs are quite clear on the usage of the UUID parameter.

florensie:
I should probably add some rendering to the test mod before this gets merged, so maybe leave this open until I get to that.

TheIllusiveC4:
> 
> 
> > I think it's useful to keep the translate methods, as different approaches without models (like rendering items or blocks) would probably appreciate it.
> 
> @emilyalexandra that's the only reason I kept them in for now, though I do still think they might need some work (if possible) as they'll only properly work with player models because of those hardcoded values.

I'm not sure those helper methods are really applicable outside of a player context if we try to abstract it out further since there's no good way to really identify where the face or leg or something of a generic entity would ideally be located.

emilyalexandra:
> 
> 
> I should probably add some rendering to the test mod before this gets merged, so maybe leave this open until I get to that.

Okay I'll wait for that, and we can keep the translate method for now, before we make an official 3.0.0 release though I may think about removing them

src/testmod/java/dev/emi/trinkets/client/TrinketsTestClient.java
```diff
@@ -0,0 +1,14 @@
...
+import dev.emi.trinkets.TrinketsTest;
+import dev.emi.trinkets.api.client.TrinketRenderer;
+import dev.emi.trinkets.api.client.TrinketRendererRegistry;
+import net.fabricmc.api.ClientModInitializer;
+
+public class TrinketsTestClient implements ClientModInitializer {
+
+	@Override
+	public void onInitializeClient() {
+		TrinketRendererRegistry.registerRenderer(TrinketsTest.TEST_TRINKET, (TrinketRenderer) TrinketsTest.TEST_TRINKET);
```
emilyalexandra, line 12: Is this cast necessary?

emilyalexandra (approved):
Looks good, I'll wait for your response and merge

src/testmod/java/dev/emi/trinkets/client/TrinketsTestClient.java
emilyalexandra: Aha, I forgot that's how items are typically stored.

PR #75, opened 2021-04-23
Add Hebrew language support

LumberWizard:
Honestly no one I know plays the game in Hebrew so no one would use this anyway

I hope the backpack translation is good but it gets the point across either way


Leo40Git:
Better translation for backpack: `"תיק גב"`

PR #87, opened 2021-05-19
Slot management refactor

TheIllusiveC4:
This PR has two main objectives:

### 1. Refactor the trinket inventory so that it's instead a collection of sub-inventories, each one representing a particular group-slot pair.

This makes it easier for modders to query particular stacks based only on the group or slot types that they're interested in, rather than needing to loop through the entire inventory.

In addition, it allows an individual slot type inventory to be resized without those changes affecting other slot type inventories. This is an important aspect that will make implementing a slot modifier system much easier (which will come in a separate PR after this one is merged).

One particular downside is that this makes looping through the entire inventory more tedious because of the presence of sub-maps. For ease of use, I added a new `forEach` method to the `TrinketComponent` interface and implement it throughout the internals for readability when we need to iterate through everything.

### 2) Refactor the rendering of slots of the same type so that they expand vertically.

This allows for more space than the current horizontal-only orientation and seems easier to keep track of slots this way. It also makes the GUI code at least 50% more cursed.

I know we've discussed having the vertical expansion hoverable as well, which should be do-able, but I figure it's fine to just have this as a baseline implementation until that's sorted out later down the line.

Example:
example

### Notes

In addition to the above objectives, I also changed the `SlotReference` object so that it holds a reference to the `TrinketInventory` it is tied to instead of the `SlotType`. I did this because the `TrinketInventory` holds much more information about the actual context of the slot, including entity information, slot type, and component reference, and it will always need to be available anyways when creating any slot reference objects.


src/main/java/dev/emi/trinkets/mixin/InventoryScreenMixin.java
```diff
@@ -157,23 +189,64 @@ private void drawForeground(MatrixStack matrices, int mouseX, int mouseY, Callba
 	@Unique
 	private void drawGroup(MatrixStack matrices, SlotGroup group) {
 		RenderSystem.enableDepthTest();
-		this.setZOffset(310);
+		this.setZOffset(305);
```
emilyalexandra, line 192: I'm curious what this change is for

emilyalexandra (approved):
Other than my comment about the one z offset change, looks good to me, this is a really serious refactor so I'm definitely going to have to test stuff out to make sure it still works in the field after the slot modifier pr.

src/main/java/dev/emi/trinkets/mixin/InventoryScreenMixin.java
TheIllusiveC4: It's just to fix this rendering issue with the slot borders overlapping the slot backgrounds:

example

Admittedly I'm not super familiar with the intricacies of the z-offsets so I'm not sure if this has any unwanted side effects, or if there's a better number for it as I chose 305 pretty arbitrarily.

PR #89, opened 2021-05-27
Slot Modifiers

TheIllusiveC4:
The second and last part of the "big PR", this part aims to add the slot modifier system to allow modders to easily and flexibly change the number of slots a player has in their Trinkets inventory.

The system mimics the vanilla entity attribute system very closely, and in fact uses `EntityAttributeModifier` for its modifier class as-is. This also means that it supports all three modifier operations: addition, base multiply, and total multiply. There may be some merit to splitting this off into its own child class, but it's currently unnecessary. However, I'll note that using it as-is in this context may be confusing since it's not technically being used on an attribute.

There are two types of modifiers, temporary and persistent. Similar to attributes, persistent modifiers are serialized and meant to be semi-permanent while temporary modifiers are purged on reconstruction. The former would be used for player progression like unlocking slots permanently and the latter would be used for items and effects that only grant the modifier when active.

There are some blocks of code that refer to the concept of "cached" modifiers. This is purely to circumvent one particular downside to this approach; items in slots granted by temporary modifiers will be ejected before those modifiers can re-applied when the entity is deserialized. The workaround is to serialize the temporary modifiers as well but remove them during the entity's first tick at the same time that modifiers are applied.

There's an additional method in the `Trinket` interface called `getSlotModifiers` that modders can use to implement simple item-based slot modifiers. Instead of an attribute, the key is a string in the format `group:slot` to represent which slot the modifier affects. The tooltip is formatted the same way as attribute modifiers and placed underneath them.

Example:
```
When equipped:
+40% Speed
+1 Ring
```


src/main/java/dev/emi/trinkets/api/Trinket.java
```diff
@@ -107,6 +107,16 @@ default boolean canUnequip(ItemStack stack, SlotReference slot, LivingEntity ent
 		return map;
 	}
 
+	/**
+	 * Returns the Slot Modifiers for a stack in a slot. Child implementations should remain pure.
+	 * Keys should be in the format "group:slot".
+	 *
+	 * @param uuid The UUID to use for creating attributes
+	 */
+	default Multimap<String, EntityAttributeModifier> getSlotModifiers(ItemStack stack, SlotReference slot, LivingEntity entity, UUID uuid) {
```
emilyalexandra, line 116: Do you think it might be beneficial to merge this and the normal modifier method and simply split them apart on the api side so it seems consistent?

emilyalexandra (approved):
Overall, seems like everything is working as it should. I have one question about merging and separating attributes, but it's definitely a curiosity, as you've worked on the system more than I. Not gonna stop this from being merged though, would still like your thoughts on it.

src/main/java/dev/emi/trinkets/api/Trinket.java
TheIllusiveC4: Hmm, that could be a good idea. I just don't know what that would look like since the maps have different key types. Would the combined method return a `Pair` of multimaps or perhaps utilize some new wrapper class? I guess it'd also be possible to convert attributes to some kind of string-key schema as well. In any case, I do think there's merit in attempting to merge them for the sake of streamlining.

emilyalexandra: Aha, that's the reason, if we go with my idea of creating a builder method, it could construct the key itself and the user would be none the wiser

PR #90, opened 2021-05-27
Remove locked status and fix 0 size

TheIllusiveC4:
Simple PR that removes the "locked" status from the slot data as that mechanic has been deprecated in favor of the slot modifiers. Also fixes the current issue with setting 0 size slot inventories, which will be important for mods to have so players can "unlock" their slots later on.


emilyalexandra:
As we talked about yesterday I just took these changes and put them into my working local branch since it has a lot of changes, closing because it's already merged on my local.

PR #93, opened 2021-06-08
Russian localization

PinkGoosik:


emilyploszaj:
Don't have anyone else who knows russian to confirm this so I ran it through deepl, just a couple questions, most translations for Trinkets continue using "Trinkets" as a name in translation strings, here it seems to translate to "keyring" (which I assume isn't a literal translation and makes more contextual sense), just want to make sure that's preferable. Also, aglet seems to have become earring, it's always funny seeing how translations handle aglet because it's a very english word, just making sure that's also correct.

PinkGoosik:
Continued in #182

PR #101, opened 2021-06-27
Fix Java versions in build.gradle

florensie:
The jitpack build is failing because it's using the wrong java version. My guess is that Jitpack read the values from the build.gradle. Changes referenced from Fabric 1.17 example mod.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,8 +4,8 @@ plugins {
 	id 'maven-publish'
 }
 
-//sourceCompatibility = JavaVersion.VERSION_1_8
-//targetCompatibility = JavaVersion.VERSION_1_8
+sourceCompatibility = JavaVersion.VERSION_16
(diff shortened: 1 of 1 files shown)


eutro:
I'm pretty sure that won't help. See this issue here: https://github.com/jitpack/jitpack.io/issues/4355.

florensie:
Ah damn, maybe it's time to set up proper maven publishing? I'm willing to offer a spot on my repository (https://maven.florens.be), and I don't doubt plenty of other community members/groups would offer access to theirs when asked.

Regardless, the changes in this PR should probably still be merged.

emilyalexandra:
Trinkets has been on a proper maven for several months (as can be noted in the readme) but you're right, still good to have in the script.

florensie:
Ah whoops, not sure how I missed that. Thanks!

PR #104, opened 2021-07-03
1.17.1

CDAGaming:


AzureDoom:
Hope this gets merged soon, as I can't work on my 1.17.1 mod anymore after updating.

gradle.properties
```diff
@@ -1,14 +1,14 @@
 # Done to increase the memory available to gradle.
-org.gradle.jvmargs=-Xmx1G
+org.gradle.jvmargs=-Xmx2G
```
emilyalexandra, line 2: What's up with this change?

emilyalexandra (commented):
Curious why a jitpack yml is included in a 1.17.1 update

gradle.properties
CDAGaming: Normally, in most projects, 2GB's is the standard in JVM args, and 1GB may not be enough for some larger development operations, so I thought to append it.

I'm happy to revert it though if there are negatives to this.

As for the jitpack file, Jitpack by default supplies only JDK 8, requiring mod devs to have that file to use JDK 16. Additionally as the TerraformersMC maven has been unreliable, I needed to use jitpack, hence prompting the file addition in case other mod devs needed to do the same.

Issue #113, opened 2021-08-01
Item in trinkets slot not showing to other players


vihangatheturtle:
Im using Give Me Hats! mod and the hats will only show for any other player on a multiplayer fabric 1.17 server if I relog. If I change the hat i am wearing i need to relog for the new hat to appear for them. Is this a trinkets issue or Give Me Hats!. There are no errors in console either when i switch hats. The 1.17.1 trinkets version doesnt work for 1.17

apace100:
I'm the author of a cosmetic armor mod which uses Trinkets in 1.17.1, and one of the players has reported the same problem with my mod (cosmetic armor only syncing when a player relogs). I came here to look at the wiki to find any information about synchronizing the Trinkets information, but couldn't find any.

I'm not sure now whether this is something that's supposed to work out of the box and is buggy, or if the wiki is just missing information.

A clarification on this would be helpful and much appreciated. :)

emilyalexandra:
as an author you should not have to do anything special, this currently seems to be either an issue with cardinal components or (more likely) an issue with how I'm using cardinal components

emilyploszaj:
Officially squashed thanks to C4's work on #119

PR #119, opened 2021-09-01
Syncing fix and miscellaneous

TheIllusiveC4:
This PR primarily fixes the syncing issue outlined here https://github.com/emilyploszaj/trinkets/issues/113.

The issue arises because we never call `ComponentKey#sync` from CCA, which is what handles the actual syncing.

However, there's a problem with using that. Because we track the Trinket changes externally (in `LivingEntityMixin`), we don't have a good way of providing the data needed for the sync if we try to use the packet writers in `AutoSyncedComponent`. Additionally, trying to use the  `readFromNbt` and `writeToNbt` implementations in `LivingEntityTrinketComponent` actually results in desyncs because those methods currently assume they're being used for entity deserialization and serialization only. Not to mention that syncing _all_ the data every time is needlessly wasteful.

Instead of trying to refactor anything, I tried to just accomplish it in the most straightforward way using our own packets/network to sync it ourselves. This way there's less room for error while keeping the amount of data being transferred in packets to a minimum.

For future reference, I think the best or cleanest solution would involve refactoring the tick method in `LivingEntityMixin` into a new API method `TrinketComponent#tick` and internally handling all of the change listeners in the component so that we can pass it onto new implementations we would write for `AutoSyncedComponent#writeSyncPacket` and `AutoSyncedComponent#applySyncPacket`. However, due to the changes and work necessary, it'd probably be best to shelve the idea until 1.18, if we even need it.

Additionally, the PR restricts both data loaders to the `trinkets` namespace so that erroneous data types do not get loaded and removes a `System.out.println` in the `SlotLoader` that I assume was left there unintentionally (let me know if this is not the case).

### DIFF
--- a/src/main/java/dev/emi/trinkets/TrinketsClient.java
+++ b/src/main/java/dev/emi/trinkets/TrinketsClient.java
@@ -6,6 +6,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import net.minecraft.util.Pair;
 import org.apache.commons.lang3.tuple.ImmutableTriple;
 import org.apache.commons.lang3.tuple.Triple;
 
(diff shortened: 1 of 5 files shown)


TheIllusiveC4:
I've just discovered a pretty rare bug, so hold off on merging this for now until I can figure out the cause. Considering how hard it is to reproduce, I can't be certain yet if it's caused by something in the PR or if it's been in for a while. Discussion on the merits of the PR itself is still open of course.

emilyploszaj:
I think I agree with your assessment that this is a bit messy, I think a 1.18 goal for a refactor in that area would be a good idea (if nothing else for getting all that code out of a mixin). My only question is does this play nicely with slot EAMs? I'm not sure if a client gets EAM information about other players and I'm worried that this might result in the number of slots not being the same for these players from the perspective of a client. Otherwise I think this is a pretty good solution.

TheIllusiveC4:
> My only question is does this play nicely with slot EAMs?

It does now. It wasn't a huge issue originally, and slot EAMs are indeed synced to other client players. However, I realized there was potential for a race condition by splitting up the packets like I did so I have combined them now into a single inventory sync packet. This ensures that all relevant slot EAMs are applied _before_ setting stacks client-side.

As for the other updates:

The packet to sync the slots data now forces an update for Trinket slots on all other client players. This should fix a small desyncing issue on clients when they receive the slots data when logging in and reloading data.

A big part of this new commit is to fix a fairly large issue with syncing slot EAMs and changing dimensions (both separately and together). Although I said we'd look into it later, I needed to override the `AutoSyncedComponent#writeSyncPacket` and `AutoSyncedComponent#applySyncPacket` methods in order to prevent CCA from calling the default `readFromNbt`, which would basically reset the slot EAMs and causes a desync. The rest of the code is just to make sure that slot EAMs are copied properly when changing dimensions, logging in, or respawning.

There's a small edit to the `PlayerScreenHandlerMixin` where I remove `trackedStacks` and `previousTrackedStacks` from `ScreenHandler` when removing slots. This is because these are added when we add slots, so should be removed also when we remove them. I'm not sure why it didn't cause any issues before, but I found that changing dimensions caused them to append extra stacks which threw a bunch of out of bounds errors (although strangely did not affect functionality).

emilyploszaj (approved):
Took a look through, everything seems to be in order, you seemed to do some pretty comprehensive tests, so I'm glad to say we can finally close this bug. Going to build a release with this later tonight.

PR #132, opened 2021-11-29
Update To 1.18

SweepSwap:


emilyploszaj (changes requested):
The 1.17 version of trinkets runs on 1.18 so I'll likely not build this until I need to add something else. There are several files with superfluous imports added for unknown reasons, can you remove those?

ProjectET:
except trinkets is not compatible with mods that use CCA 4.0.0.

PR #141, opened 2022-01-07
Create pt_br.json (Brazilian Portuguese)

ppblitto:


src/main/resources/assets/trinkets/lang/pt_br.json
```diff
@@ -0,0 +1,20 @@
...
+	"trinkets.slot.chest.necklace": "Colar",
+	"trinkets.slot.chest.back": "Costas",
+	"trinkets.slot.chest.cape": "Capa",
+	"trinkets.slot.legs.belt": "Cinto",
+	"trinkets.slot.feet.aglet": "Agulheta",
+	"trinkets.slot.feet.shoes": "Pés",
+	"trinkets.slot.hand.glove": "Mão principal",
+	"trinkets.slot.offhand.glove": "Mão secundária",
+	"trinkets.slot.hand.ring": "Anel da mão principal",
+	"trinkets.slot.offhand.ring": "Anel da mão secundária"
```
emilyploszaj, line 19: in english the glove and ring slots are named the same thing, Trinkets does extra handling to treat slots named the same as the same thing when displaying it to a player, so something equippable in both slots would say "Equippable in Ring slot" instead of listing both primary and offhand. is there a language restriction from simply calling these "glove" and "ring"?

emilyploszaj (commented):
also, a lot of these translations seem to use body parts instead of accessory name (for example, feet instead of shoes). is there a language reason that makes this more comfortable to phrase it as?

PR #154, opened 2022-02-27
Update for 1.18.2

florensie:
Only issue so far appears to be getting the list of entity types from an entity type tag in [EntitySlotLoader](https://github.com/florensie/trinkets/blob/d3b86d60d62bdf8f97920456f1ba4c76c83801f8/src/main/java/dev/emi/trinkets/data/EntitySlotLoader.java#L140).

One possible solution is to do something like vanilla recipe loading. Create abstract entries (Ingredient.Entry <- StackEntry, TagEntry), which can then be lazily loaded after resource loading is complete.	

~Another solution would be to call getRegistryTags on TagManagerLoader but it appears that getting the instance of that class might be non-trivial.~

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 plugins {
-	id 'fabric-loom' version '0.10-SNAPSHOT'
-	//id "com.github.fudge.forgedflowerloom" version "2.0.0"
+	id 'fabric-loom' version '0.11-SNAPSHOT'
+	id 'io.github.juuxel.loom-quiltflower' version '1.6.0'
 	id 'maven-publish'
 }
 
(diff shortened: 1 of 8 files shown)


Azazelthedemonlord:
Hey, not sure if you have seen this yet, but I feel this may be handy for ya'll.
https://gist.github.com/apple502j/f20dd720ca4fc292a87c7ad9dbed8dfa 

https://gist.github.com/apple502j/4d4e57f067f5be408556001086949434 

https://gist.github.com/apple502j/4f0eaabb1d178cc92f0940b40361933b

florensie:
Thanks! Of note from the first link:

> To get DataPackContents, you need to use accessor to access MinecraftServer#resourceManagerHolder, use an access widener to make MinecraftServer$ResourceManagerHolder public, and finally call dataPackContents() - however in most cases there are shortcut methods in MinecraftServer. 

I had already found this as a way to get TagManagerLoader, which would contain the tags we need, but as I said we need an instance of the server. I could always try to cheese it with `FabricLoader.getInstance().getGameInstance()` and some careful casting.

emilyploszaj:
I've gotten around to porting but this is still listed as a draft. Is it ready to merge or are you going to continue working on it?

florensie:
Aside from the issue explained above, it's ready to merge. I have some idea of how to get around the issue but it might take me a while until I get around to implementing it. You're free to do it yourself, I explained it in the description.

Right now you could:
- Merge and accept the issue for now, I don't think it's a feature that is used by many
- Fix it yourself: merge immediately or fix it on this PR branch, you should have push rights
- Wait until I have time to fix the issue

PR #161, opened 2022-04-07
Slot fixes

TheIllusiveC4:
This is a PR discussed briefly over Discord before. In summary, this PR addresses two fairly niche slot issues.

Firstly, default slots get all of its default predicates overwritten if any mod attempts to implement additional predicates. This is due to the fact that the default predicates are only assigned if the predicates are empty to begin with. 

https://github.com/emilyploszaj/trinkets/blob/366f80866ad978b1204785333c6d210b1dd9d786/src/main/java/dev/emi/trinkets/data/SlotLoader.java#L142-L150

Related issue from my Trinket Shulker Boxes mod: https://github.com/TheIllusiveC4/CuriousShulkerBoxes/issues/16

While there are many potential solutions, it was decided that the easiest and most straightforward method was to explicitly define the default predicates in the provided default slot json files.

Secondly, slot modifiers will not accurately persist across loads. This causes items in slots given by modifiers to be placed back into the player's inventory upon reloading into a world despite the fact that the slots still seem to exist. This occurs because although the modifiers are correctly saved and loaded, they are cleared and then re-applied in the first tick.

https://github.com/emilyploszaj/trinkets/blob/366f80866ad978b1204785333c6d210b1dd9d786/src/main/java/dev/emi/trinkets/mixin/LivingEntityMixin.java#L143

My proposed solution involves moving this logic to the end of `LivingEntityTrinketComponent#readFromNbt`. Once the initial data is read and loaded, all of the slot modifiers from the worn items are read and then subsequently removed from the modifiers enqueued for removal. Essentially, this is saying "these are the slot modifiers we had before, here are the items we have now, keep the slot modifiers from those items and remove the rest, then proceed with first tick." Doing it this way also ensures that items that have changed their slot modifiers due to a mod update will propagate those changes correctly.

I tested the first change with Inmis and Trinket Shulker Boxes and it does correctly solve the issue so that items from both mods are equippable again when loaded together. I tested the second change with the Trinkets test mod: I wore the item to give more ring slots, added diamonds as a ring item, and then placed diamonds in all of the ring slots. Relogging keeps the diamonds in place. I also tested both changes with Botania loaded to make sure it didn't somehow adversely affect other trinkets.

### DIFF
--- a/src/main/java/dev/emi/trinkets/api/LivingEntityTrinketComponent.java
+++ b/src/main/java/dev/emi/trinkets/api/LivingEntityTrinketComponent.java
@@ -7,6 +7,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.UUID;
 import java.util.function.BiConsumer;
 import java.util.function.Predicate;
 
@@ -17,6 +18,7 @@ import dev.emi.trinkets.TrinketPlayerScreenHandler;
 import dev.onyxstudios.cca.api.v3.component.sync.AutoSyncedComponent;
 import net.fabricmc.fabric.api.util.NbtType;
 import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.attribute.EntityAttribute;
 import net.minecraft.entity.attribute.EntityAttributeModifier;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.item.ItemStack;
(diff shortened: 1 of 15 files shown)


emilyploszaj (changes requested):
I believe we discussed this and my mind was on validators. I'm not sure we want `trinkets:all` on the quick move and tooltip predicates, though. That means they *have* to be overwritten to have any effect at all to reduce their effect.

TheIllusiveC4:
Oh right, very good point. That slipped my mind, there's no real reason those need to be there.

PR #183, opened 2022-06-07
1.19 port

Nyfaria:
straight port, no changes


emilyploszaj:
Hi, there was an open 1.19 port PR that had a few more features than this by a long term contributor, so I merged it instead, thanks for your contribution though!

PR #194, opened 2022-07-25
[1.19]: Fix for Crash within SinglePlayer during the /reload command

### LINKED ISSUE #175: [1.18.2 Fabric] Opening the Creative Inventory menu after doing /reload with KubeJS crashes the game.
I'm making my own modpack and I wanted to use KubeJS to add in custom textures and recipes, however after using /reload to test out my new recipes, my game crashed. I even took off any mods that added trinket slots to the GUI and it still didn't help.
When I tested it without KubeJS, the problem pretty much went away.

Here is a crashlog as well as a video of the bug:
https://crashy.net/GyKKbO4OQSmlEdbERVq2
https://drive.google.com/file/d/1l2wqYVgc7wT5ZC9Xa1v_jwPcHo0Mmwgq/view?usp=sharing

Dragon-Seeker:
Fixes a bug where for whatever reason within a single player world, the client player's screen handler will have data within the maps but during the /reload, new SlotGroup's are made which a very different instances compared to the ones within the given maps causing null values to be passed causing a crash if the CreativeScreen or Inventory Screen is open.

Should fix #175 

Pr Adds:
- Comparison Methods to the SlotGroup class
- Force Redraw of the Players Inventory Screen if opened

### DIFF
--- a/src/main/java/dev/emi/trinkets/TrinketPlayerScreenHandler.java
+++ b/src/main/java/dev/emi/trinkets/TrinketPlayerScreenHandler.java
@@ -33,8 +33,8 @@ public interface TrinketPlayerScreenHandler {
 
 	int trinkets$getTrinketSlotEnd();
 
-	boolean shouldReRenderScreen();
+	boolean trinkets$shouldReRenderScreen();
 
-	void setReRenderScreen(boolean reRender);
+	void trinkets$setReRenderScreen(boolean reRender);
 
 }
--- a/src/main/java/dev/emi/trinkets/TrinketPlayerScreenHandler.java
+++ b/src/main/java/dev/emi/trinkets/TrinketPlayerScreenHandler.java
@@ -1,9 +1,11 @@
 package dev.emi.trinkets;
 
 import java.util.List;
+import java.util.Map;
 
 import dev.emi.trinkets.api.SlotGroup;
 import dev.emi.trinkets.api.SlotType;
+import net.minecraft.screen.slot.Slot;
 
 /**
  * Interface for putting methods onto the player's screen handler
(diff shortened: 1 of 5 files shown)


emilyploszaj (changes requested):
There are a couple problems with this PR. First of all, there are names in interfaces put on vanilla classes that don't follow the standard of prefixing with mod name, this can cause incompatibilities, especially with "shouldReRenderScreen", seems likely to conflict relatively speaking.
I can't tell why a hashcode and equality was provided for slot groups, it doesn't seem to be used? Reference equality is intentional for slot groups, and there are a lot of fields not accounted for in the comparison.
Lastly, the core functionality of this seems to be just, disabling the screen until the handler decides it's safe which is... peculiar. It doesn't seem like a proper solution to this problem, and could probably be prone to breakage. A more adequate solution would probably involve figuring out why there's a desync between consensus of the handler and screen, and correcting it at the source. Since you're only disabling *rendering* here, interaction can almost certainly still NPE.

Dragon-Seeker:
I am not "Disabling" screen, just forcing it to redraw with the correctly loaded data after a /reload due to some internal Reference mismatch. It could be due the the syncing implementation but this fixes a critical crash during /reload

Dragon-Seeker:
The hashcode and equals method is used internally to find if a given value is contained within the map meaning it is better to compare the values internally rather than the Reference equality.

emilyploszaj:
But reference equality is what's desired, slots can change during reload. The fact that they're recreated is intentional, and I'm sure they're used in maps in other places where reference equality is important, this is a big breaking change that I haven't had the time to check the significance of.

Dragon-Seeker:
You are right, the problem comes down to a singleton issue with both the Server and the Client sharing the EntitySlotLoader.INSTANCE variable due to running on the same JVM. Meaning the when the Server updates the EntityType map for slots, the client is instantly updated but the components are yet to update causing a delay allowing for atleast a frame to be rendered leading to the Null Pointer Exception - (Found with help from glisco)

PR #195, opened 2022-07-26
[1.19]: Fix for Crash within SinglePlayer during the /reload command with NPE

Dragon-Seeker:
Part two: Electric Boogaloo

Explanation: 

The NPE comes from a Singleton issue with both the Server and the Client sharing the EntitySlotLoader.INSTANCE variable due to running on the same JVM in Single Player World. Meaning the when the Server updates the EntityType map for slots, the client is instantly updated but the components are yet to update causing a delay allowing for atleast a frame to be rendered leading to the Null Pointer Exception

Without the Help from @gliscowo  I would have never been able to understand the underline problem.


Dragon-Seeker:
A possible re-write to the API could fix this by not using the a Static Instance of EntitySlotLoader and maybe putting the slot data fully with the Cardinal Component's themself but this is a bandaid that stops the crash from the NPE

Issue #215, opened 2022-09-20
[Feature Request] Add Trinket#getMaxCount(ItemStack stack, SlotReference slot, LivingEntity entity)


Sollace:
Currently all trinket slots are hardcoded to take the maximum stack count for whatever item you put into it. I have items for which I'd actually like these to function like armour slots, but not without losing the ability to stack the item when storing it normally.

i.e. In the case of a glove, though you can stack it up to 64 in a chest, it doesn't make sense to be able to wear more than one on each hand.

emilyploszaj:
This is pretty significantly difficult to implement due to certain vanilla systems. Shift clicking in particular would basically have to be removed in order to properly support items that vary in stack size based on slot.

Sollace:
Um... No?

Armour slots already do this. If I implemented it would you accept a pull request?

emilyploszaj:
You are incorrect, armor slots do not support what you are suggesting. Shift clicking ignores slot max size, and this is the reason that beacons don't support shift clicking, and why brewing stands, the enchantment table, and the player inventory have very specific, custom (sometimes broken) logic for shift clicking. You can easily display a fault by placing a bunch of stacks of 1 glass bottle in the inventory and shift clicking them into a brewing stand, despite the slot only supporting a single item, they'll stack up to 64. You can even use this to get stacks larger than 64 in your cursor.

The reason you can shift click carved pumpkins onto your head is because the vanilla inventory hardcodes a solution for them, it only attempts normal shift click behavior if the slot is empty (this is non standard). This is fine, because `transferSlot` works for moving items to empty slots okay, but breaks when combining. This means trinkets could only ever support a max stack size of 1. Simple experiment, set a slot's max count to 2, place a single item in it, and shift click 64 of the item in your inventory, the limited slot will have a full stack in it.

Remedying this properly is non-trivial. Invasively patching the vanilla method is prone to breakage of expectations for other modders beyond what can be anticipated. Reimplementing the method can cause broken compatibility with other mods, and since trinkets is messing with the quick moving of the vanilla inventory, I do not want to poke that.

A more fundamental API problem is that vanilla slots do not support contextual max counts. That is, you cannot have a slot that can hold 2 apples, or 4 sticks. You can only set a max count for the slot. Trying to work around this restriction statefully in hardcoded interactions is extremely prone to breakage.

[...]

PR #216, opened 2022-09-21
Implement per-slot stack limits 

Sollace:
Resolves https://github.com/emilyploszaj/trinkets/issues/215

I took into account your concerns added some assurances to the mixin:
1. I set its priority (to 1002) to run after most other mods have injected, so if there are conflicts other mods shouldn't fail to inject due to the presence of this mod
2. I've set the require to 0 for the injections in this mod, so that in the unlikely scenario that there are conflicts, this mod's patches will simply not be applied. Since this only affects transferring items to trinket slots, I figured a minor bug caused by a conflict here (rather than a crash) is tolerable.
3. I've added a check so that vanilla behaviour is preserved. Out custom logic only comes into effect when trying to insert stacks into slots added by trinkets.

The new method in Trinket is `Tinket#getMaxCount(ItemStack stack, SlotReference slot)` The only difference here from the opened issue is removing the entity argument as that would have complicated things further, and appears not all that neccessary for the method to be useful anyway since you're already given both the stack and the slot.

```
	/**
	 * Determines the maximum number of this item that can be inserted into a particular trinket slot.
	 *
	 * @param stack The stack being equipped
	 * @param slot The slot the stack is being equipped into
	 * @return The maximum stack size for the given slot
	 *
	 * @since 3.4.1
	 */
	default int getMaxCount(ItemStack stack, SlotReference slot) {
		return stack.getMaxCount();
	}
```

Relevant changes to use this method are in `TrinketItem#equipItem` to make it aware of stack limits, and `CreativeTrinketSlot` and `SurvivalTrinketSlot` where I've overriden `Slot#getMaxItemCount(ItemStack stack)` to return the result of the above method.

Finally for testing I've added a `stackable_test` item to the test mod that can be normally stacked to 2 but when placed in the hat slot only stacks 1. 

On initial test, clicking the item into the slot normally only deposits one item from the stack into the slot, and it does not accept any more after that. When shift-clicking it will transfer one item from the stack to the slot, and any remaining will be deposited elsewhere.

The only unintended, though arguably sensible behaviour I noticed, is that if you have more than one valid slot an item can go into, shift-click will deposit one item into each available slot rather than just one into the first.


Sollace:
> dozens of injection points if you're counting more than annotations
This is three redirects. First one targets two locations, second targets two locations, and the third targets one location (notice ordinal = 0) for a total of 5 injection points. 

> especially considering how much people often want to mess with screens. 
From my experience very few people inject into this method.

> quickly spot an NPE waiting to happen if some but not all of the redirects fail.
Which NPE please. Are you talking about `mustChangeBehaviour(currentSlot)`? instanceof checks return false for null values. There is no chance of a NPE there. If you see any other places where a NPE might occur, it would be useful if you could point them out.

> Priority 1002 is weird, you really should do a larger number like 2000 so people can actually order stuff properly
I figured 1002 would be suffient, but sure, I'll increase that.

> getMaxCount probably should've included an entity parameter for API method parity
See my initial comment for why an entity parameter isn't included (and isn't needed).

> Version incrementation is not the responsibility of feature PRs
Fair. You will need to increment the version and update the docs' @since annotation when you release though.

>  Put a single item in a slot, shift click a stack, they'll merge above the slot limit.
I tested (and recently retested) this case and it does not occur with the patches as they are.

Sollace:
Tested again with a item stack limit of 64 and a slot stack limit of 2. No perceived change. You cannot add more than 2 items into the trinket slot when it it limited to 2, neither by shift-clicking nor my clicking directly on the slot.

image

Likewise, you cannot swap in more than 2 of the item that introduces the limit when there is a different item in the slot.

image

Dispensers didn't want to insert into trinket slots, so I couldn't test that.

Sollace:
> you'd expect the slot to fill and leave 1 behind, instead it'll skip, or equip to some other slot, which is frustrating and unexpected behavior.

This is something I mentioned in my initial post. Check the last paragraph where I described the behaviour. I agree, it is somewhat frustrating an unexpected.

> Overall, as a fundamental API in the Fabric space, so I care a lot about Trinkets' stability. These redirects, including a no op redirect to capture a local, are touching a lot of areas in this method, and I am not comfortable with risking incompat for a relatively niche use case. I'd probably be fine with this in a content mod, or something that had significantly fewer users, but Trinkets is included in basically every modpack released which means I'm rolling the dice for incompats on basically every mod that has been or will be created. I've already submitted bug reports to Mojang about the implementation of this method a while ago, and that is my ideal route to getting something like this implemented.

That's a reasonable concern, I understand. Perhaps my tolerance for multiple injections is a bit higher because in general I don't I have to consider cases like there. I do agree, and hope Mojang does fix the method themselves in a later case.

If this isn't going to be merged, I'll understand. In that case I hope you don't mind that I stick to using the changes I worked out personally for my own mod where I do need this.

Issue #98, opened 2021-06-19
Mending elytra not repairing in trinket slot


FakeDomi:
When using the Elytra Trinkets mod, mending elytras cannot repair in the cape slot.

FakeDomi:
Fixed it myself now, feel free to pull the fix in if you want to.

PinkGoosik:
I think it should be merged into elytra trinket mod and not trinkets

emilyalexandra:
Mending is a vanilla behavior and should be working through trinkets itself, it worked in 1.16 but I forgot about  it when I rewrote it for 1.17, it'll get done the next chance I get

emilyalexandra:
Fixed in 3.0.1

PinkGoosik:
pog

yesn0t:
hi, mending elytra still not working most of the time for me, by that i mean XP gained rarely ever goes to mend the elytra;  
i did tests with XP bottles : if i splash 1 xp bottle, while a damaged mending elytra is in trinket slot, then two problems occur :  
* at least 4 out of 5 times the elytra will not be mended
* when it does get mended, the durability gain is 6 points instead of ~18

with elytra in chestplate slot, mending works normally :  
* 100% of the time, elytra is mended by incoming XP
* each mend gives a lot more than 6 points of durability

using `elytratrinket 3.0.0` and `trinkets 3.2.0`  

please let me know if you can reproduce this issue or is if it's just me (._.)'

FakeDomi:
Experiencing the same issue in 3.4.0 still @yesn0t @emilyploszaj

Issue #220, opened 2022-11-19
Item with mending in cape slot doesn't get repaired


jonalange:
I use the Armored Eyltra from [Better End](https://www.curseforge.com/minecraft/mc-mods/betterend) in the cape trinket slot.

I enchanted it with Mending and Unbreaking, but when receiving XP the eyltra in the slot doesn't get repaired.

When I hold it in the hand it repairs normally as it should.

So maybe this is a compatibility issue of both mods together or this mod has a bug.
Because the elytra repairs itself in the hand or off hand slot, I think the latter is the case.

If you need any further information, let me know :)

PR #222, opened 2022-11-20
Fix Mending mixin

TheIllusiveC4:
A fairly straightforward fix for Mending enchantment checks in Trinket slots. Addresses https://github.com/emilyploszaj/trinkets/issues/220 and some reports in the closed https://github.com/emilyploszaj/trinkets/issues/98.

The issue is that the `Enchantments.MENDING.getEquipment` call is used as if it returns all equipped Mending items when it actually just returns all equipped items regardless of enchantment. This causes the behavior to appear correct if a player only has Mending equipment, but misfires proportionally to how many non-Mending equipment they have.

The solution is to just filter for the enchantment after the call but before calculating the total number of candidates. The code is re-arranged a bit just so the same predicate can be used for both the Trinket slots and the regular slots.

There's also a null check in there because the initial entry being modified can be null if no initial candidate was found from the vanilla slots, allowing us to just skip that segment entirely in that case.

I tested this in two different instances using Trinkets and Elytra Slot. While wearing a Mending elytra in the Cape slot and using Experience Bottles, the elytra will always be repaired if I leave the other equipment slots empty. Donning a shield and armor (all without Mending) reduces the chance of this happening significantly. With this fix applied, the elytra will always be repaired in either scenario.

### DIFF
--- a/src/main/java/dev/emi/trinkets/mixin/ExperienceOrbEntityMixin.java
+++ b/src/main/java/dev/emi/trinkets/mixin/ExperienceOrbEntityMixin.java
@@ -1,5 +1,6 @@
 package dev.emi.trinkets.mixin;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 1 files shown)


emilyploszaj (approved):

jonalange:
Is this already in the current version? Because the release notes say absolutely nothing I'm not sure about it.

PR #265, opened 2023-07-15
Create zh_cn.json

RookiesTwo:


src/main/resources/assets/trinkets/lang/zh_cn.json
```diff
@@ -0,0 +1,20 @@
+{
+	"trinkets.tooltip.slots.any" : "可装备于 §9any§r 饰品槽",
```
emilyploszaj, line 2: Why leave "any" untranslated here? Is there is a word or perhaps another phrase that would be more suitable?

RookiesTwo: sorry, that was a mistake

emilyploszaj (approved):

PR #269, opened 2023-07-25
German Translation

BoyPika:


emilyploszaj (commented):
Just a curiosity, is the word chosen here for slot the same as what minecraft would generally use? I also believe line 2 has a grammatical error.

BoyPika:
The translators for Minecraft seem to avoid using the word "slot." For example, in the controls menu the "hotbar slot" text was replaced with "Schnellzugriff" or quick access. As for the grammatical error, I believe that I've now fixed it.

PR #292, opened 2024-03-04
Tick trinkets on ```LivingEntity``` instances

powerboat9:
This should tick trinkets on non-player living entities as well as player entities


emilyploszaj:
What are the differences between both instances on players? Does it tick properly once per tick on both client and server in both cases? Do living entities tick on client? Does this occur notably earlier or later in a normal tick call? I'm not super comfortable with that question mark in a comment not seeming to know what's going on but I think this is a reasonable change if you've looked into these questions.

powerboat9:
```tick``` calls ```tickMovement``` which calls ```this.inventory.updateItems```. The question mark is there because while I've preserved the order of ticking trinkets and updating the client, I don't know if its important. Note that ```tickMovement``` is only called on entities which haven't been removed, which I've just corrected

emilyploszaj:
And the status on client/server ticking parity? Have you done logging on a test trinket to make sure everything gets ticked as often as it should on both sides?

TheIllusiveC4:
Some preliminary testing on my end show that client/server ticking parity is as expected (once per tick on each side) with this change.

emilyploszaj:
Then it's reasonable enough to me.

PR #293, opened 2024-03-04
Modify ```equipItem``` to accept ```LivingEntity```

powerboat9:


emilyploszaj (changes requested):
This is a breaking API changes that would crash countless mods.

powerboat9:
Sorry, created a new method ```equipItemLiving``` instead

emilyploszaj:
Is it possible to have them be named the same?

powerboat9:
I wasn't sure, but it looks like [yes](https://docs.oracle.com/javase/specs/jls/se9/html/jls-8.html#jls-8.4.9)

emilyploszaj (approved):

Issue #272, opened 2023-08-12
Intigration with REI menu windows


Shibva:
![image](https://github.com/emilyploszaj/trinkets/assets/38770780/6b5fd91c-b127-4ee8-9173-0a1770f5953e)
![image](https://github.com/emilyploszaj/trinkets/assets/38770780/1fb6c8a5-ed77-48ed-b94f-5cbbc550748f)

There are mays that these types of windows can be shifted out of the way. It would be nice if this can be inpimented due to the window itself if you trying to equip or unequip one will cause the element to be activated

Shibva:
Yay; thankyou!

PR #295, opened 2024-03-05
Add REI and EMI exclusion areas for compatibility

TheIllusiveC4:
Addresses https://github.com/emilyploszaj/trinkets/issues/272.

Currently, there is an issue where Trinket slots may block or overlap elements of REI/EMI resulting in unintended interactions. This PR adds exclusion areas for REI and EMI compatibility using their respective plugin APIs, which allows for elements of those recipe viewers to shift out of the way of Trinket slots when appropriate. This has been tested in both Survival and Creative inventory screens.

JEI compatibility was also attempted, and would be trivial to add, but I could not figure a resolution to the mapping issue that occurs when using a Mojmap dependency in a Yarn environment.

Cloth Config is part of the compile dependencies due to it packaging the `Rectangle` class needed for REI exclusion areas.


emilyploszaj (approved):
It might be a minor performance improvement to not specify exclusion areas within the screen bounds itself... but that's probably negligible. Re: JEI, that's totally fine on Fabric, JEI is basically unused. I'd frankly be fine with only EMI compat but since the REI compat is there no reason to toss it away. Shame about all the dependencies though.

Issue #253, opened 2023-06-06
Trinkets onEquip and onUnequip is not called when player swaps trinket


Tiviacz1337:
Fabric 1.19.2 - 0.14.19
Trinkets 3.4.0
Methods Trinket#onEquip and Trinket#onUnequip are not called if player swaps trinket that is the same item, but with different nbt data.

emilyploszaj:
This behavior was changed, the methods are now called.

PR #296, opened 2024-03-05
Adjust equip/unequip behavior

### LINKED ISSUE #217: onEquip/onUnequip not run with Trinkets differentiated by Enchantment
The oldStack -> newStack check in LivingEntity's `tick` doesn't capture my trinkets being enchantment-driven. I have 3 common trinkets but 20-odd enchantments that can go on them that define their functionality.

[this bit](https://github.com/emilyploszaj/trinkets/blob/8b234a847c1abab65b6b52ff4e8ae086f8bc8c57/src/main/java/dev/emi/trinkets/mixin/LivingEntityMixin.java#L196):
```java
if (!newStack.isItemEqual(oldStack)) {
    TrinketsApi.getTrinket(oldStack.getItem()).onUnequip(oldStack, ref, entity);
    TrinketsApi.getTrinket(newStack.getItem()).onEquip(newStack, ref, entity);
}
```

One headband/whatever looks like any other of that trinket according to this, but the enchantments may be different, and that's not captured/calling onEquip/onUnequip when you swap my trinkets because of this. you have to remove old trinket to inventory then equip new one to the empty slot.

This also may cause issues if some trinket, for example, had a durability associated with it. Any functionally-important NBT really.

`ItemStack.areEqual(oldStack,newStack)` may resolve some of the issues mentioned.

emilyploszaj:
Equip/unequip now is called if stacks of the same item are changed at all, not just if items change, also smartly ignores mutation that occurs during ticking.

Fixes #217

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -100,7 +100,7 @@ jar {
 publishing {
 	publications {
 		mavenJava(MavenPublication) {
-            artifact(sourcesJar) {
+            artifact(remapSourcesJar) {
                 builtBy remapSourcesJar
             }
             afterEvaluate {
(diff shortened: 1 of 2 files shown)


TheIllusiveC4 (approved):
Looks good to me. This should also resolve https://github.com/emilyploszaj/trinkets/issues/253.

Issue #237, opened 2023-03-17
Creative players cannot unequip items with curse of binding from trinket slots


Sollace:
The normal behaviour in vanilla is that players in creative mode can still unequip armour even when it has curse of binding.

It seems to be an oversight in Trinket#canUnequip where the method does not check if the wearer is a creative mode player.

```
	/**
	 * Determines whether an entity can unequip a trinket
	 *
	 * @param stack The stack being unequipped
	 * @param slot The slot the stack is being unequipped from
	 * @param entity The entity that is unequipping the stack
	 * @return Whether the stack can be unequipped
	 */
	default boolean canUnequip(ItemStack stack, SlotReference slot, LivingEntity entity) {
		return !EnchantmentHelper.hasBindingCurse(stack);
	}
```

This is the fix I'm implementing in my subclass:

```
    @Override
    public boolean canUnequip(ItemStack stack, SlotReference slot, LivingEntity entity) {
        return !(EnchantmentHelper.hasBindingCurse(stack) && EntityPredicates.EXCEPT_CREATIVE_OR_SPECTATOR.test(entity));
    }
```

And a reference to vanilla to show that this is the intended behaviour:

![image](https://user-images.githubusercontent.com/6429283/226003455-619d729c-6e73-4d49-999a-256d17e79cf0.png)

Issue #206, opened 2022-08-17
Curse of Binding Trinkets cannot be removed in creative mode


Modulations:
As the title says, even though you can remove curse of binding armor in creative, you can't remove curse of binding trinkets in creative.

Rav115:
since 
trinkets-3.4.2.jar 1.19.2
trinkets-3.5.1.jar 1.19.3
we have 
![Screenshot_20230116_070551](https://user-images.githubusercontent.com/39887530/212602672-e098e318-1588-480b-babf-3acc33ddc69e.png)
**/trinkets set offhand ring 0** air or **/execute as @p run trinkets set head hat 0 air**
and using it can remove item with curse of binding even if in creative so kinda fixed
sadly still dont have item modify to remove just the enchantment

PR #299, opened 2024-03-25
Allow Curse of Binding trinkets to be removed by Creative players

TheIllusiveC4:
Mostly self-explanatory. Addresses https://github.com/emilyploszaj/trinkets/issues/206 and https://github.com/emilyploszaj/trinkets/issues/237.


emilyploszaj (approved):
Starting to wonder if this check should even be in the base Trinket case. I suppose it is valuable to be able to intentionally disable it. Either way, this looks reasonable.

PR #304, opened 2024-04-05
Add `canEquipFromUse` and `getEquipSound` methods to `Trinket` interface

TheIllusiveC4:
This PR aims to add two new API methods to the `Trinket` interface.

### `canEquipFromUse`

Currently, developers that want to implement "equip from use" behavior needs to implement it themselves. This is convenient to do in most cases through the `TrinketItem.equipItem` static method, which can be used from anywhere, including an item's `use` implementation. However, certain situations make this method less feasible. If a developer does not own the item, as is the case for mods that add Trinkets compatibility to vanilla/other mods, they would need to call this method from an event listener that they register themselves. In addition, mods that want to keep their Trinkets compatibility modular (i.e. multi-loader or optional dependents) will have a trickier time accommodating a static method outside of the interface.

This PR adds a `canEquipFromUse` method to the `Trinket` interface that is called from an internal `UseItemCallback` event listener. The method defaults to `false` so existing behavior won't be changed for any items. If it's overridden by an implementation to be `true`, the `TrinketItem.equipItem` method is called for the stack and subsequent processing is canceled. This precludes other use behavior on the item, but any item that already has a use behavior should not have need of this in the first place.

### `getEquipSound`

The only reference for trinket equip sounds is in `TrinketItem.equipItem`, where it defers back to the vanilla `Equipment` interface. Developers may want to provide their own equip sounds for their trinkets without the side effects of implementing the `Equipment` interface, as well as allowing different sounds based on slot or stack data. This PR adds the `getEquipSound` method to the `Trinket` interface and replaces it in the call inside `TrinketItem.equipItem`. The default for the method defers back to the `Equipment` interface as before so existing behavior should not be changed due to this.

### DIFF
--- a/src/main/java/dev/emi/trinkets/TrinketsMain.java
+++ b/src/main/java/dev/emi/trinkets/TrinketsMain.java
@@ -5,6 +5,11 @@ import static com.mojang.brigadier.arguments.StringArgumentType.string;
 import static net.minecraft.server.command.CommandManager.argument;
 import static net.minecraft.server.command.CommandManager.literal;
 
+import dev.emi.trinkets.api.Trinket;
+import dev.emi.trinkets.api.TrinketItem;
+import net.fabricmc.fabric.api.event.player.UseItemCallback;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.TypedActionResult;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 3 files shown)


emilyploszaj (approved):
This looks reasonable. It's worth noting that a mod is able to mixin to the default implementation of Trinket and every single item will attempt to equip itself, but that's not really an error, just an implementation detail.

Issue #307, opened 2024-04-27
Update to 1.20.5


NyakoFox:
I'm not opening this as a "please do this urgently" type of thing, instead just as a way to mark it's not compatible yet.

vico93:
I second this! Also i take the opportunity to inform that 1.20.6 almost 100% compatible with 1.20.5 mods, since the only change for now is fixing [MC-271109](https://bugs.mojang.com/browse/MC-271109)

Ioandar:
yes please! 
I have a mod on rollout testing for fabric 1.20.5 and would love to include my trinkets anklet :D

rakantoh:
+1 to update, thanks

PR #310, opened 2024-05-08
Update to 1.20.6

Patbox:
This pr updates trinkets to version 1.20.6, with minimal api changes to make it match with how MC handles things (mostly moving from EntityAttribute to RegistryEntry<>, as MC requires them everywhere) and to update the dependencies.

Additionally it adds basic support for datafixers updating ItemStacks stored in trinket's inventory (this is unlikely to require any changes in the future).

Solves #307


emilyploszaj (changes requested):
The codebase doesn't use star imports and uses tabs instead of spaces. At least one large file has replaced all tabs with spaces making the diff completely illegible, and the added files use spaces instead of tabs, this will need to be resolved before being properly reviewed.

Patbox:
Done

HumanGamer:
I just tried this 1.20.6 version and noticed an issue. With keep inventory enabled, on death, trinkets get deleted. I tried to debug this but couldn't figure out why. Not sure if this is an issue outside of this PR but I've only tried on 1.20.6 with this.

Patbox:
Should be fixed now

src/main/java/dev/emi/trinkets/mixin/LivingEntityMixin.java
```diff
@@ -52,6 +53,7 @@
  *
  * @author Emi
  */
+@SuppressWarnings("UnreachableCode")
```
emilyploszaj, line 56: SuppressWarnings also unneeded here.

src/main/java/dev/emi/trinkets/mixin/LivingEntityMixin.java
```diff
@@ -163,37 +165,52 @@ private void tick(CallbackInfo info) {
...
 						if (!oldStack.isEmpty()) {
 							Trinket trinket = TrinketsApi.getTrinket(oldStack.getItem());
-							Multimap<EntityAttribute, EntityAttributeModifier> map = trinket.getModifiers(oldStack, ref, entity, uuid);
+							Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> map = trinket.getModifiers(oldStack, ref, entity, uuid);
 							Multimap<String, EntityAttributeModifier> slotMap = HashMultimap.create();
-							Set<SlotEntityAttribute> toRemove = Sets.newHashSet();
-							for (EntityAttribute attr : map.keySet()) {
-								if (attr instanceof SlotEntityAttribute slotAttr) {
+							Set<RegistryEntry<EntityAttribute>> toRemove = Sets.newHashSet();
+							for (var attr : map.keySet()) {
```
emilyploszaj, line 171: Don't use `var`.

src/main/java/dev/emi/trinkets/mixin/LivingEntityMixin.java
```diff
@@ -212,29 +229,19 @@ private void tick(CallbackInfo info) {
...
 
 				if (!contentUpdates.isEmpty() || !inventoriesToSend.isEmpty()) {
-					PacketByteBuf buf = PacketByteBufs.create();
-					buf.writeInt(entity.getId());
-					NbtCompound tag = new NbtCompound();
+					var map = new HashMap<String, NbtCompound>();
 
 					for (TrinketInventory trinketInventory : inventoriesToSend) {
-						tag.put(trinketInventory.getSlotType().getGroup() + "/" + trinketInventory.getSlotType().getName(), trinketInventory.getSyncTag());
+						map.put(trinketInventory.getSlotType().getGroup() + '/' + trinketInventory.getSlotType().getName(), trinketInventory.getSyncTag());
```
emilyploszaj, line 235: Not that important but why the change from a string to a char here? I preferred the string.

src/main/java/dev/emi/trinkets/mixin/datafixer/Schema1460Mixin.java
```diff
@@ -0,0 +1,34 @@
...
+import org.spongepowered.asm.mixin.Unique;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import java.util.Map;
+import java.util.function.Supplier;
+
+@Mixin(Schema1460.class)
+public class Schema1460Mixin {
```
emilyploszaj, line 19: I said it in discord but I do not care for this, I can't easily understand this part of the codebase, all of the methods are unmapped, and the benefit is pretty low level that I've never supported before, I'd prefer this class be removed.

src/main/resources/trinkets.mixins.json
```diff
@@ -3,23 +3,24 @@
...
   "compatibilityLevel": "JAVA_17",
   "mixins": [
-    "accessor.ScreenHandlerAccessor",
+    "ExperienceOrbEntityMixin",
     "LivingEntityMixin",
     "PlayerManagerMixin",
     "PlayerScreenHandlerMixin",
     "ServerPlayNetworkHandlerMixin",
-    "ExperienceOrbEntityMixin"
+    "accessor.ScreenHandlerAccessor",
```
emilyploszaj, line 11: Don't adjust the order of the mixins here, subpackages (`accessor`) go first.

emilyploszaj (changes requested):
Functionality wise I imagine it works and looks good in that sense. There were a lot of crits here, mostly style stuff, I don't know if I caught them all but my codebases don't use star imports or `var`, `info` is always used instead of `ci` or `cir`, and the handful of spot specific things I mentioned. Some comments need further clarification.

src/main/java/dev/emi/trinkets/api/TrinketsAttributeModifiersComponent.java
```diff
@@ -0,0 +1,98 @@
...
+import net.minecraft.entity.attribute.EntityAttribute;
+import net.minecraft.entity.attribute.EntityAttributeModifier;
+import net.minecraft.network.RegistryByteBuf;
+import net.minecraft.network.codec.PacketCodec;
+import net.minecraft.network.codec.PacketCodecs;
+import net.minecraft.registry.Registries;
+import net.minecraft.registry.RegistryKeys;
+import net.minecraft.registry.entry.RegistryEntry;
+
+public record TrinketsAttributeModifiersComponent(List<Entry> modifiers, boolean showInTooltip) {
```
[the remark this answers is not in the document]
Patbox: It replaces reading attributes from nbt, as ItemStacks use components now (also as a bonus, it works like an api extension as it can be set on items by default)

src/main/java/dev/emi/trinkets/mixin/LivingEntityMixin.java
Patbox: Added it purely for dev, as my ide marked all code post casting as unreachable, but can remove that

src/main/resources/trinkets.mixins.json
Patbox: Going to blame the ide here, but going to fix that

src/main/java/dev/emi/trinkets/mixin/datafixer/Schema1460Mixin.java
Patbox: I, as a developer/user of this library prefer for Trinkets to support this, as losing things like item data/enchantments isn't fun. In cases where people use trinkets purely for things as elytra cape slot will likely want to not lose the data.

The only thing happening here is adding nbt structure of trinkets data to definitions of player and entity. It doesn't do anything else, as vanilla handles the rest. These methods are unnamed, as they are just lambdas, which never have a name in yarn.

src/main/java/dev/emi/trinkets/api/Trinket.java
```diff
@@ -145,7 +131,7 @@ default Multimap<EntityAttribute, EntityAttributeModifier> getModifiers(ItemStac
 
 	/**
 	 * Called by Trinkets when a trinket is broken on the client if {@link TrinketsApi#onTrinketBroken}
-	 * is called by the consumer in {@link ItemStack#damage(int, LivingEntity, Consumer)} server side
+	 * is called by the consumer in {@link ItemStack#damage(int, LivingEntity, EquipmentSlot)} server side
```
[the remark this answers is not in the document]
Patbox: Yes it got removed

src/main/java/dev/emi/trinkets/api/Trinket.java
```diff
@@ -113,30 +115,14 @@ default SoundEvent getEquipSound(ItemStack stack, SlotReference slot, LivingEnti
...
-						EntityAttributeModifier entityAttributeModifier = EntityAttributeModifier.fromNbt(tag);
-
-						if (entityAttributeModifier != null
-								&& entityAttributeModifier.getId().getLeastSignificantBits() != 0L
-								&& entityAttributeModifier.getId().getMostSignificantBits() != 0L) {
-							map.put(optional.get(), entityAttributeModifier);
-						}
-					}
+	default Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> getModifiers(ItemStack stack,
+																						   SlotReference slot, LivingEntity entity, UUID uuid) {
```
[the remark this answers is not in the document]
Patbox: Mostly written that on a laptop and it just didn't fit on the screen. But cqn fix that

src/main/java/dev/emi/trinkets/mixin/ClientPlayNetworkHandlerMixin.java
```diff
@@ -25,7 +26,7 @@
 public class ClientPlayNetworkHandlerMixin {
 
     @Inject(at = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerEntity;setId(I)V"), method = "onPlayerRespawn", locals = LocalCapture.CAPTURE_FAILSOFT)
-    private void onPlayerRespawn(PlayerRespawnS2CPacket packet, CallbackInfo info, CommonPlayerSpawnInfo commonPlayerSpawnInfo, RegistryKey<World> registryKey, RegistryEntry<DimensionType> registryEntry, ClientPlayerEntity clientPlayerEntity, ClientPlayerEntity clientPlayerEntity2)  {
+    private void onPlayerRespawn(PlayerRespawnS2CPacket packet, CallbackInfo ci, CommonPlayerSpawnInfo commonPlayerSpawnInfo, RegistryKey registryKey, RegistryEntry registryEntry, ClientPlayerEntity clientPlayerEntity, RegistryKey registryKey2, boolean bl, DownloadingTerrainScreen.WorldEntryReason worldEntryReason, ClientPlayerEntity clientPlayerEntity2)  {
```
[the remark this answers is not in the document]
Patbox: Mostly because it was done by ide instead of by hand, through going to replace it with @Local from MixinExtras instead

Patbox:
Fixed all the style issues.

I'm keeping datafixer, but commented on what every part of it does

Lakuna:
Hello @Patbox,

I updated one of my mods to 1.20.6 using this fork and found that Trinkets is now required on the client side. This appears to be related to Ladysnake/Cardinal-Components-API#174, which is addressed in the migration guide [here](https://ladysnake.org/wiki/cardinal-components-api/upgrade-instructions/CCA-6-changes#client-optional-components). It would be nice if this could be fixed before this fork is merged (it appears to work fine otherwise).

Thanks!

Patbox:
@Lakuna Could you tell me which case you would need that? Trinkets (at least the original version) never really worked with vanilla/non-trinkets environments, as it's not possible to interact with these slots at all. It can also be confusing, as trinkets changes shift clicking behaviour to put items within it's slot to add support for it. Additionally it wouldn't work anyway, as TrinketsAttributeModifiersComponent, which was added to replicate attribute support in nbt, but also can be used for defining the attributes in Item.Settings is required to be on both sides

Lakuna:
> @Lakuna Could you tell me which case you would need that? Trinkets (at least the original version) never really worked with vanilla/non-trinkets environments, as it's not possible to interact with these slots at all. It can also be confusing, as trinkets changes shift clicking behaviour to put items within it's slot to add support for it. Additionally it wouldn't work anyway, as TrinketsAttributeModifiersComponent, which was added to replicate attribute support in nbt, but also can be used for defining the attributes in Item.Settings is required to be on both sides

My server was previously able to be connected to by vanilla clients who didn't wish to use trinkets or by clients with the mod who did want to use trinkets. The only mod that I use that "adds" a trinket is one that allows Elytra to be used in the cape slot. Since this update, players that belong to the former group are no longer able to connect. It would be strange to me if this update would impose a restriction that didn't exist beforehand.

Patbox:
Going to leave that decision for @emilyploszaj . For myself I will update the polymer port for server side usage

rakantoh:
sorry if i an noob, but where i can  download the commits/forks made by @Patbox? Thanks.

rakantoh (approved):

emilyploszaj (approved):
Docs still say "by the consumer" when a consumer is no longer present, I've left the issues open. As far as I can tell every other crit is addressed, and the PR can be merged after this.

Patbox:
Not sure how I missed that one again, but now it's changed

PR #319, opened 2024-05-29
Update to 1.21

Patbox:
Updates trinkets to 1.21 with a set of breaking changes™

Currently a draft as it's unfinished, untested AND crashes due to Cardinal Component dependency not being updated yet


Patbox:
Got in game and it worked without issues. Ready for review!

src/main/java/dev/emi/trinkets/api/SlotType.java
```diff
@@ -156,4 +156,8 @@ public boolean equals(Object o) {
 	public int hashCode() {
 		return Objects.hash(group, name);
 	}
+
+	public String getId() {
```
emilyploszaj, line 159: Why is this getting added to API?

src/main/java/dev/emi/trinkets/api/SlotAttributes.java
```diff
@@ -1,32 +1,32 @@
...
+	private static final Map<String, Identifier> CACHED_IDS = Maps.newHashMap();
+	private static final Map<String, RegistryEntry<EntityAttribute>> CACHED_ATTRIBUTES = Maps.newHashMap();
 	
 	/**
 	 * Adds an Entity Attribute Nodifier for slot count to the provided multimap
 	 */
-	public static void addSlotModifier(Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> map, String slot, UUID uuid, double amount,
-			EntityAttributeModifier.Operation operation) {
+	public static void addSlotModifier(Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> map, String slot, Identifier identifier, double amount,
+									   EntityAttributeModifier.Operation operation) {
```
emilyploszaj, line 21: This indentation shouldn't be added.

src/main/resources/trinkets.mixins.json
```diff
@@ -3,24 +3,25 @@
...
-    "datafixer.Schema1460Mixin",
-    "ExperienceOrbEntityMixin",
+    "EnchantmentDefinitionMixin",
+    "EnchantmentHelperMixin",
     "LivingEntityMixin",
     "PlayerManagerMixin",
     "PlayerScreenHandlerMixin",
-    "ServerPlayNetworkHandlerMixin"
+    "ServerPlayNetworkHandlerMixin",
+    "accessor.ScreenHandlerAccessor",
```
emilyploszaj, line 12: Don't move the package mixins.

emilyploszaj (changes requested):
Looks mostly fine, some crits though. Why did the mending mixin disappear?

src/main/java/dev/emi/trinkets/api/SlotType.java
Patbox: Since it is used in a lot of places (and now more due to vanilla enchantment changes) I though it made sense to do this. But I can remove that if you prefer.

Patbox:
It is handled now by EnchantmentHelperMixin as "repair_with_xp" functionality uses the EnchantmentHelper.chooseEquipmentWith(...) method to get the item. The new mixin actually should cover way more cases than original implementation, as it allows protection-alike enchantments for trinkets to work.

src/main/java/dev/emi/trinkets/api/SlotType.java
emilyploszaj: Nah you're right this is good to have. Weird to call it an ID when it's not an identifier, but I'll live.

emilyploszaj (approved):
Looks good to me. Are there any outstanding issues to look at before I release this?

Patbox:
Only issue I can think of would be change of format for EntityAttributeModifier's nbt (due to name + uuid -> identifier) change, but that should only result on few warns and persistant/cached modifiers being cleared.

Issue #309, opened 2024-05-07
[Suggestion] Add equip/unequip event


ochotonida:
Currently, the only way to detect whether an item is being unequipped or equipped in a trinket slot is to have a Trinket attached to the itemstack that overrides `onUnequip`, or to manually keep track of all equipped items.
When adding effects based on an equipped item's components/nbt, attaching a Trinket to the stack might not be ideal or possible if the item is not owned by the mod. Adding an event similar to Curios' CurioChangedEvent, that fires any time an equipped item changes, would resolve this.

Issue #289, opened 2024-01-08
clearing trinket slots


LordMaximus05:
Is there a way to clear trinket slots with commands yet?

PR #320, opened 2024-06-15
Equip/Unequip Events; Polish of Trinket Attribute Modifiers

fzzyhmstrs:
## Resolves:
* https://github.com/emilyploszaj/trinkets/issues/309
* https://github.com/emilyploszaj/trinkets/issues/289
* Trinkets with multiple "same" slots (hand/ring, offhand/ring) still showing multiple repeating attribute blocks
* Trinket slots with multiple slots having modifier collisions

## Explanation
👍 Added new `TrinketEquipCallback` and `TrinketUnequipCallback`. Each have the same parameters as the `Trinket` method itself. They are wired in the same spot as the Trinket method, directly after the trinkets method is called. Added simple test registrations for them; both working.

👍Added new method `getRefId()` to SlotReference; tucking away the messy inventory().slotType().getId() etc etc.

👍Used this new method in a new method in `SlotAttributes`; `toSlotReferencedModifier`. This method suffixes the modifier with the `getRefId()`. This avoids collisions if multiple slots are available for one "slot", or if trinkets could go into multiple/any slot, and attributes are being added via NBT. Previously the identifier added via NBT would be used directly; causing conflicts if 2+ of the same NBT trinket are equipped.

👍Incorporated this new method into a new static method in `Trinket`; `trinketModifiers`. This flips the script on the previous `Build NBT Attributes -> Hope the user calls super -> return map`. Now the flow is `Provide blank multimap if the user calls super -> user adds code-side modifiers -> Trinket adds NBT modifiers -> return map`. This flow is tested in the new TestTrinket2, which does NOT call super, and yet both modifiers are still apparent (presuming you `/give` yourself a TestTrinket2 with some NBT modifiers).
image

Also encapsulates `getTrinket` and `SlotAttributes.getIdentifier`, as at every Trinkets call site those were simply generated immediately before passing them in. Now they are just generated in situ, with an overload available to be custom if needed.

Technically a implementation change in the off chance that someone was relying on finding an NBT attribute to determine if they do anything attribute related in code (despite your insistence in the docs to keep the contract pure). Unsure the impact compared to people just not calling super (definitely happens). Understandable if you want to flip this logic back, with the other improvements intact.

👍Wired the new `trinketModifiers` method into all the Trinkets call sites.

👍Polished `ItemStackMixin`: 
* Tooltip hiding is checked up front and short circuits the method if all relevant trinket tooltips are hidden anyways.
* Fixed "same" slots still being shown separately. Easy exmaple `hand/ring` and `offhand/ring`. Reason this was still happening: Equality of `SlotType` was checked with the translation, but the relevant maps were then stored with the `SlotType` as the key, undoing this check. Map now uses the `MutableText` that was checked as the key directly. Results:

Before:
image

After (Test Trinket 2 is equippable in both ring slots):
image

👍Added `clear` argument to the trinkets command (`/trinkets clear`). This clears out all of the trinket slots.

## Could Do
Could add `canEquip` and `canUnequip` callbacks to have Event functionality for those main 4 trinket methods. I stopped at `onEquip` and `onUnequip` because that was the asked content in the issue, and custom predicates can be made. 

## Recommend
Adding info about the two new events to the Wiki

### DIFF
--- a/src/main/java/dev/emi/trinkets/TrinketModifiers.java
+++ b/src/main/java/dev/emi/trinkets/TrinketModifiers.java
@@ -1,7 +1,11 @@
 package dev.emi.trinkets;
 
 import com.google.common.collect.Multimap;
-import dev.emi.trinkets.api.*;
+import dev.emi.trinkets.api.SlotAttributes;
+import dev.emi.trinkets.api.SlotReference;
+import dev.emi.trinkets.api.Trinket;
+import dev.emi.trinkets.api.TrinketsApi;
+import dev.emi.trinkets.api.TrinketsAttributeModifiersComponent;
 import net.minecraft.entity.LivingEntity;
 import net.minecraft.entity.attribute.EntityAttribute;
 import net.minecraft.entity.attribute.EntityAttributeModifier;
@@ -11,33 +15,27 @@ import net.minecraft.util.Identifier;
 
 public class TrinketModifiers {
 
-    //internalizes getTrinket and slotIdentifier, both which typically are generated just before the modifiers call anyway
-    public static Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> get(ItemStack stack, SlotReference slot, LivingEntity entity){
-        Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> map = TrinketsApi.getTrinket(stack.getItem()).getModifiers(stack, slot, entity, SlotAttributes.getIdentifier(slot));
-        if (stack.contains(TrinketsAttributeModifiersComponent.TYPE)) {
-            for (TrinketsAttributeModifiersComponent. Entry entry : stack.getOrDefault(TrinketsAttributeModifiersComponent.TYPE, TrinketsAttributeModifiersComponent.DEFAULT).modifiers()) {
-                if (entry.slot().isEmpty() || entry.slot().get().equals(slot.inventory().getSlotType().getId())) {
-                    map.put(entry.attribute(), toSlotReferencedModifier(entry.modifier(), slot));
-                }
-            }
-        }
-        return map;
-    }
+	//internalizes getTrinket and slotIdentifier, both which typically are generated just before the modifiers call anyway
+	public static Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> get(ItemStack stack, SlotReference slot, LivingEntity entity){
+		Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> map = TrinketsApi.getTrinket(stack.getItem()).getModifiers(stack, slot, entity, SlotAttributes.getIdentifier(slot));
+		if (stack.contains(TrinketsAttributeModifiersComponent.TYPE)) {
+			for (TrinketsAttributeModifiersComponent. Entry entry : stack.getOrDefault(TrinketsAttributeModifiersComponent.TYPE, TrinketsAttributeModifiersComponent.DEFAULT).modifiers()) {
+				map.put(entry.attribute(), entry.modifier());
+			}
+		}
+		return map;
+	}
 
-    //overload if a custom method for retrieving the trinket is used. Also exposes the slotIdentifier if custom on that is needed
-    public static Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> get(Trinket trinket, ItemStack stack, SlotReference slot, LivingEntity entity, Identifier slotIdentifier){
-        Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> map = trinket.getModifiers(stack, slot, entity, slotIdentifier);
-        if (stack.contains(TrinketsAttributeModifiersComponent.TYPE)) {
-            for (TrinketsAttributeModifiersComponent. Entry entry : stack.getOrDefault(TrinketsAttributeModifiersComponent.TYPE, TrinketsAttributeModifiersComponent.DEFAULT).modifiers()) {
-                if (entry.slot().isEmpty() || entry.slot().get().equals(slot.inventory().getSlotType().getId())) {
-                    map.put(entry.attribute(), toSlotReferencedModifier(entry.modifier(), slot));
-                }
-            }
-        }
-        return map;
-    }
-
-    public static EntityAttributeModifier toSlotReferencedModifier(EntityAttributeModifier modifier, SlotReference ref){
-        return new EntityAttributeModifier(modifier.id().withSuffixedPath("/" + ref.getId()), modifier.value(), modifier.operation());
-    }
+	//overload if a custom method for retrieving the trinket is used. Also exposes the slotIdentifier if custom on that is needed
+	public static Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> get(Trinket trinket, ItemStack stack, SlotReference slot, LivingEntity entity, Identifier slotIdentifier){
+		Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> map = trinket.getModifiers(stack, slot, entity, slotIdentifier);
+		if (stack.contains(TrinketsAttributeModifiersComponent.TYPE)) {
+			for (TrinketsAttributeModifiersComponent. Entry entry : stack.getOrDefault(TrinketsAttributeModifiersComponent.TYPE, TrinketsAttributeModifiersComponent.DEFAULT).modifiers()) {
+				if (entry.slot().isEmpty() || entry.slot().get().equals(slot.inventory().getSlotType().getId())) {
+					map.put(entry.attribute(), entry.modifier());
+				}
+			}
+		}
+		return map;
+	}
 }
--- a/src/main/java/dev/emi/trinkets/TrinketModifiers.java
+++ b/src/main/java/dev/emi/trinkets/TrinketModifiers.java
@@ -1,0 +1,43 @@
+package dev.emi.trinkets;
+
+import com.google.common.collect.Multimap;
+import dev.emi.trinkets.api.*;
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.attribute.EntityAttribute;
+import net.minecraft.entity.attribute.EntityAttributeModifier;
+import net.minecraft.item.ItemStack;
+import net.minecraft.registry.entry.RegistryEntry;
+import net.minecraft.util.Identifier;
+
+public class TrinketModifiers {
+
+    //internalizes getTrinket and slotIdentifier, both which typically are generated just before the modifiers call anyway
+    public static Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> get(ItemStack stack, SlotReference slot, LivingEntity entity){
+        Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> map = TrinketsApi.getTrinket(stack.getItem()).getModifiers(stack, slot, entity, SlotAttributes.getIdentifier(slot));
+        if (stack.contains(TrinketsAttributeModifiersComponent.TYPE)) {
+            for (TrinketsAttributeModifiersComponent. Entry entry : stack.getOrDefault(TrinketsAttributeModifiersComponent.TYPE, TrinketsAttributeModifiersComponent.DEFAULT).modifiers()) {
+                if (entry.slot().isEmpty() || entry.slot().get().equals(slot.inventory().getSlotType().getId())) {
+                    map.put(entry.attribute(), toSlotReferencedModifier(entry.modifier(), slot));
+                }
+            }
+        }
+        return map;
+    }
+
+    //overload if a custom method for retrieving the trinket is used. Also exposes the slotIdentifier if custom on that is needed
+    public static Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> get(Trinket trinket, ItemStack stack, SlotReference slot, LivingEntity entity, Identifier slotIdentifier){
+        Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> map = trinket.getModifiers(stack, slot, entity, slotIdentifier);
+        if (stack.contains(TrinketsAttributeModifiersComponent.TYPE)) {
+            for (TrinketsAttributeModifiersComponent. Entry entry : stack.getOrDefault(TrinketsAttributeModifiersComponent.TYPE, TrinketsAttributeModifiersComponent.DEFAULT).modifiers()) {
+                if (entry.slot().isEmpty() || entry.slot().get().equals(slot.inventory().getSlotType().getId())) {
+                    map.put(entry.attribute(), toSlotReferencedModifier(entry.modifier(), slot));
+                }
+            }
+        }
+        return map;
+    }
+
+    public static EntityAttributeModifier toSlotReferencedModifier(EntityAttributeModifier modifier, SlotReference ref){
+        return new EntityAttributeModifier(modifier.id().withSuffixedPath("/" + ref.getId()), modifier.value(), modifier.operation());
+    }
+}
--- a/src/main/java/dev/emi/trinkets/TrinketsMain.java
+++ b/src/main/java/dev/emi/trinkets/TrinketsMain.java
@@ -106,7 +106,31 @@ public class TrinketsMain implements ModInitializer, EntityComponentInitializer 
 						)
 					)
 				)
+				.then(
+					literal("clear")
+					.executes(context -> {
+						try {
+							return clearCommand(context);
+						} catch (Exception e){
+							e.printStackTrace();
+							return -1;
+						}
+					})
+				)
 			));
+	}
+
+	private static int clearCommand(CommandContext<ServerCommandSource> context){
+		ServerPlayerEntity player = context.getSource().getPlayer();
+		if (player != null) {
+			TrinketComponent comp = TrinketsApi.getTrinketComponent(player).get();
+			for (var entry : comp.getInventory().entrySet()){
+				for (var inv : entry.getValue().values()){
+					inv.clear();
+				}
+			}
+		}
+		return 1;
 	}
 
 	private static int trinketsCommand(CommandContext<ServerCommandSource> context, int amount) {
--- a/src/main/java/dev/emi/trinkets/TrinketsMain.java
+++ b/src/main/java/dev/emi/trinkets/TrinketsMain.java
@@ -5,14 +5,7 @@ import static com.mojang.brigadier.arguments.StringArgumentType.string;
 import static net.minecraft.server.command.CommandManager.argument;
 import static net.minecraft.server.command.CommandManager.literal;
 
-import dev.emi.trinkets.api.Trinket;
-import dev.emi.trinkets.api.TrinketItem;
-import dev.emi.trinkets.api.TrinketComponent;
-import dev.emi.trinkets.api.LivingEntityTrinketComponent;
-import dev.emi.trinkets.api.TrinketsApi;
-import dev.emi.trinkets.api.TrinketsAttributeModifiersComponent;
-import dev.emi.trinkets.api.SlotGroup;
-import dev.emi.trinkets.api.SlotType;
+import dev.emi.trinkets.api.*;
 import dev.emi.trinkets.payload.BreakPayload;
 import dev.emi.trinkets.payload.SyncInventoryPayload;
 import dev.emi.trinkets.payload.SyncSlotsPayload;
@@ -45,6 +38,9 @@ import net.minecraft.resource.ResourceType;
 import net.minecraft.server.command.ServerCommandSource;
 import net.minecraft.server.network.ServerPlayerEntity;
 import net.minecraft.text.Text;
+import org.spongepowered.asm.mixin.injection.struct.InjectorGroupInfo;
+
+import java.util.Map;
 
 public class TrinketsMain implements ModInitializer, EntityComponentInitializer {
 
@@ -124,8 +120,8 @@ public class TrinketsMain implements ModInitializer, EntityComponentInitializer 
 		ServerPlayerEntity player = context.getSource().getPlayer();
 		if (player != null) {
 			TrinketComponent comp = TrinketsApi.getTrinketComponent(player).get();
-			for (var entry : comp.getInventory().entrySet()){
-				for (var inv : entry.getValue().values()){
+			for (Map.Entry<String, Map<String, TrinketInventory>> entry : comp.getInventory().entrySet()){
+				for (TrinketInventory inv : entry.getValue().values()){
 					inv.clear();
 				}
 			}
--- a/src/main/java/dev/emi/trinkets/api/SlotAttributes.java
+++ b/src/main/java/dev/emi/trinkets/api/SlotAttributes.java
@@ -28,10 +28,6 @@ public class SlotAttributes {
 		return CACHED_IDS.computeIfAbsent(key, Identifier::of);
 	}
 
-	public static EntityAttributeModifier toSlotReferencedModifier(EntityAttributeModifier modifier, SlotReference ref){
-		return new EntityAttributeModifier(modifier.id().withSuffixedPath("/" + ref.getId()), modifier.value(), modifier.operation());
-	}
-
 	public static class SlotEntityAttribute extends EntityAttribute {
 		public String slot;
 
--- a/src/main/java/dev/emi/trinkets/api/SlotAttributes.java
+++ b/src/main/java/dev/emi/trinkets/api/SlotAttributes.java
@@ -13,9 +13,9 @@ import net.minecraft.util.Identifier;
 public class SlotAttributes {
 	private static final Map<String, Identifier> CACHED_IDS = Maps.newHashMap();
 	private static final Map<String, RegistryEntry<EntityAttribute>> CACHED_ATTRIBUTES = Maps.newHashMap();
-	
+
 	/**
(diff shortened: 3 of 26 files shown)


src/main/java/dev/emi/trinkets/TrinketsMain.java
```diff
@@ -106,9 +106,33 @@ public void onInitialize() {
...
+					})
+				)
 			));
 	}
 
+	private static int clearCommand(CommandContext<ServerCommandSource> context){
+		ServerPlayerEntity player = context.getSource().getPlayer();
+		if (player != null) {
+			TrinketComponent comp = TrinketsApi.getTrinketComponent(player).get();
+			for (var entry : comp.getInventory().entrySet()){
```
emilyploszaj, line 127: No `var` please, makes it hard to review and maintain. I'll note this in further comments as well

src/main/java/dev/emi/trinkets/api/SlotAttributes.java
```diff
@@ -24,17 +24,20 @@ public static void addSlotModifier(Multimap<RegistryEntry<EntityAttribute>, Enti
...
 	public static Identifier getIdentifier(SlotReference ref) {
-		String key = ref.inventory().getSlotType().getId() + "/" + ref.index();
-		CACHED_IDS.computeIfAbsent(key, Identifier::of);
-		return CACHED_IDS.get(key);
+		String key = ref.getRefId();
+		return CACHED_IDS.computeIfAbsent(key, Identifier::of);
+	}
+
+	public static EntityAttributeModifier toSlotReferencedModifier(EntityAttributeModifier modifier, SlotReference ref){
+		return new EntityAttributeModifier(modifier.id().withSuffixedPath("/" + ref.getRefId()),modifier.value(),modifier.operation());
```
emilyploszaj, line 32: Please put spaces between args. What is the use case of this method?

src/main/java/dev/emi/trinkets/api/SlotReference.java
```diff
@@ -1,4 +1,7 @@
 package dev.emi.trinkets.api;
 
 public record SlotReference(TrinketInventory inventory, int index) {
+    public String getRefId() {
```
emilyploszaj, line 4: Blank line before methods please, also I feel like this should probably also be `getId`, what do you think?

src/main/java/dev/emi/trinkets/api/Trinket.java
```diff
@@ -103,26 +103,43 @@ default RegistryEntry<SoundEvent> getEquipSound(ItemStack stack, SlotReference s
 		return stack.getItem() instanceof Equipment eq ? eq.getEquipSound() : null;
 	}
 
+	//internalizes getTrinket and slotIdentifier, both which typically are generated just before the modifiers call anyway
+	static Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> trinketModifiers(ItemStack stack, SlotReference slot, LivingEntity entity){
```
emilyploszaj, line 107: I don't really understand why these exist. You cannot overload static methods. Additionally, the styling and documentation is inadequate. I feel like this should remain baked in behavior of the bottom level getModifiers

src/main/java/dev/emi/trinkets/api/Trinket.java
```diff
@@ -103,26 +103,43 @@ default RegistryEntry<SoundEvent> getEquipSound(ItemStack stack, SlotReference s
...
+		}
+		return map;
+	}
+
 	/**
 	 * Returns the Entity Attribute Modifiers for a stack in a slot. Child implementations should
 	 * remain pure
 	 * <p>
-	 * If modifiers do not change based on stack, slot, or entity, caching based on passed UUID
+	 * If modifiers do not change based on stack, slot, or entity, caching based on passed slotIdentifier
```
emilyploszaj, line 136: slotIdentifier -> Identifier

src/main/java/dev/emi/trinkets/api/event/TrinketEquipCallback.java
```diff
@@ -0,0 +1,25 @@
...
+
+import dev.emi.trinkets.api.SlotReference;
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ItemStack;
+
+public interface TrinketEquipCallback {
+    Event<TrinketEquipCallback> EVENT = EventFactory.createArrayBacked(TrinketEquipCallback.class,
+    listeners -> (stack, slot, entity) -> {
```
emilyploszaj, line 11: This entire subsection needs to be indented further

src/main/java/dev/emi/trinkets/api/event/TrinketEquipCallback.java
```diff
@@ -0,0 +1,25 @@
...
+        for (var listener: listeners){
+            listener.onEquip(stack,slot,entity);
+        }
+    });
+
+    /**
+     * Called when an entity equips a trinket, after the `onEquip` method of the Trinket
```
emilyploszaj, line 18: \`onEquip\` isn't adequate javadocs, consider using @link

src/main/java/dev/emi/trinkets/api/event/TrinketUnequipCallback.java
```diff
@@ -0,0 +1,25 @@
+package dev.emi.trinkets.api.event;
+
+import dev.emi.trinkets.api.SlotReference;
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ItemStack;
+
+public interface TrinketUnequipCallback {
```
emilyploszaj, line 9: Same notes here as equip callback, indent the lambda and document the link properly. additionally, "unequip" should be preferred over "un-equip"

src/main/java/dev/emi/trinkets/mixin/ItemStackMixin.java
```diff
@@ -49,9 +50,16 @@ public abstract class ItemStackMixin {
 	private void getTooltip(Item.TooltipContext context, PlayerEntity player, TooltipType tooltipType, CallbackInfoReturnable<List<Text>> info, @Local(ordinal = 0) List<Text> list) {
 		TrinketsApi.getTrinketComponent(player).ifPresent(comp -> {
 			ItemStack self = (ItemStack) (Object) this;
+
+			boolean hideAdditionalTooltip = self.contains(DataComponentTypes.HIDE_ADDITIONAL_TOOLTIP);
+			boolean showAttributeTooltip = self.getOrDefault(DataComponentTypes.ATTRIBUTE_MODIFIERS, AttributeModifiersComponent.DEFAULT).showInTooltip();
+			if (hideAdditionalTooltip && !showAttributeTooltip)
```
emilyploszaj, line 56: Do not use 1 line if statements.

src/main/java/dev/emi/trinkets/mixin/ItemStackMixin.java
```diff
@@ -49,9 +50,16 @@ public abstract class ItemStackMixin {
...
+				//literally nothing to do
+				return;
+
 			boolean canEquipAnywhere = true;
 			Set<SlotType> slots = Sets.newHashSet();
-			Map<SlotType, Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier>> modifiers = Maps.newHashMap();
+			Map<MutableText, Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier>> modifiers = Maps.newHashMap();
```
emilyploszaj, line 62: I believe I prefer the original implementation here, what are the benefits to having mutable text be the key rather than a potentially sortable slot type?

src/main/java/dev/emi/trinkets/mixin/LivingEntityMixin.java
```diff
@@ -62,12 +64,12 @@
 public abstract class LivingEntityMixin extends Entity {
 	@Unique
 	private final Map<String, ItemStack> lastEquippedTrinkets = new HashMap<>();
-	
+
 	@Shadow
-	protected abstract AttributeContainer getAttributes();
+	public abstract AttributeContainer getAttributes();
 
-	private LivingEntityMixin() {
```
emilyploszaj, line 69: Revert this constructor change.

src/main/java/dev/emi/trinkets/mixin/LivingEntityMixin.java
```diff
@@ -128,10 +130,12 @@ private void dropInventory(CallbackInfo info) {
 		}));
 	}
 
+	@Unique
 	private void dropFromEntity(ItemStack stack) {
 		ItemEntity entity = dropStack(stack);
 		// Mimic player drop behavior for only players
-		if (entity != null && ((Entity) this) instanceof PlayerEntity) {
+        //noinspection ConstantValue
```
emilyploszaj, line 137: I prefer to not use noinspection in my codebase

src/testmod/java/dev/emi/trinkets/TestTrinket2.java
```diff
@@ -0,0 +1,46 @@
...
+public class TestTrinket2 extends TrinketItem {
+
+	public TestTrinket2(Settings settings) {
+		super(settings);
+	}
+
+	@Override
+	public Multimap<RegistryEntry<EntityAttribute>, EntityAttributeModifier> getModifiers(ItemStack stack, SlotReference slot, LivingEntity entity, Identifier id) {
+		/*//testing the composition of the new attribute suffix
+		TrinketsTest.LOGGER.info(SlotAttributes.toSlotReferencedModifier(new EntityAttributeModifier(id.withSuffixedPath("trinkets-testmod/movement_speed"),
```
emilyploszaj, line 38: This probably shouldn't be commented out, right?

src/testmod/java/dev/emi/trinkets/TrinketsTest.java
```diff
@@ -13,14 +17,26 @@ public class TrinketsTest implements ModInitializer {
...
+	public static Item TEST_TRINKET_2;
 
 	@Override
 	public void onInitialize() {
 		LOGGER.info("[Trinkets Testmod] test mod was initialized!");
 		TEST_TRINKET = Registry.register(Registries.ITEM, identifier("test"), new TestTrinket(new Item.Settings().maxCount(1).maxDamage(100)));
+		TEST_TRINKET_2 = Registry.register(Registries.ITEM, identifier("test2"), new TestTrinket2(new Item.Settings().maxCount(1)));
+		TrinketEquipCallback.EVENT.register(((stack, slot, entity) -> {
+			if(stack.isOf(TEST_TRINKET_2)){
+				entity.getWorld().playSound(null,entity.getBlockPos(), SoundEvents.ENTITY_ARROW_HIT, SoundCategory.PLAYERS, 1f,1f);
```
emilyploszaj, line 29: Please have spaces between args here and below.

emilyploszaj (changes requested):
Lots of style crits here and I remain unconvinced about some of the changes you've made, particularly in `Trinket`. Overall I think the events are nice and I can see the value in some of the other changes but I'd like to hear some responses. I agree with passing a blank multimap from super and providing nbt specific modifiers from elsewhere but this should not be exposed as API, because static methods cannot be overwritten.

src/main/java/dev/emi/trinkets/api/SlotReference.java
fzzyhmstrs: getId is fine, just slightly confusing with a getId in a related, contained thing. But fine.

src/main/java/dev/emi/trinkets/api/SlotAttributes.java
fzzyhmstrs: Spaxes can be added. Explained in the original comment. Provides differentiation between modifiers made via nbt that otherwise all use the same nbt-provided identifier and collide when used in multiple places.

src/main/java/dev/emi/trinkets/TrinketsMain.java
fzzyhmstrs: Can remove. Its not the only var in your codebase fyi. Can track the others down too.

src/main/java/dev/emi/trinkets/api/event/TrinketEquipCallback.java
fzzyhmstrs: Thought I matched the indenting of the Drop callback exactly. Will check

src/main/java/dev/emi/trinkets/mixin/ItemStackMixin.java
fzzyhmstrs: Explained in the original comment. This resolves your slot "collapsing" not working

src/testmod/java/dev/emi/trinkets/TestTrinket2.java
fzzyhmstrs: Super spams the log if left in. Can leave it in if you really prefer

src/main/java/dev/emi/trinkets/api/Trinket.java
fzzyhmstrs: Ok

src/main/java/dev/emi/trinkets/api/event/TrinketUnequipCallback.java
fzzyhmstrs: Same reply

src/main/java/dev/emi/trinkets/mixin/ItemStackMixin.java
fzzyhmstrs: If sortability is truly a concern, the underlying text string be used.

src/main/java/dev/emi/trinkets/mixin/LivingEntityMixin.java
fzzyhmstrs: Ok

src/main/java/dev/emi/trinkets/mixin/LivingEntityMixin.java
fzzyhmstrs: Ok

src/testmod/java/dev/emi/trinkets/TrinketsTest.java
fzzyhmstrs: Ok

src/main/java/dev/emi/trinkets/api/Trinket.java
fzzyhmstrs: If visibility is the concern, I can move them elsewhere. Probably best you tell we where so I'm not just hunting for the proper spot myself

src/main/java/dev/emi/trinkets/mixin/ItemStackMixin.java
fzzyhmstrs: Ok

fzzyhmstrs:
Also general comment re: `var`. You allowed several through previous review, so this critique is a fix of code previous to my work.

src/main/java/dev/emi/trinkets/mixin/ItemStackMixin.java
fzzyhmstrs: Actually on.. fourth? Look, my original explanation doesn't seem to be correct. The actual fix comes from the change in the equality checking below this. I'll have to take (fifth) look tomorrow. The justification I have now is 'reduced cpu usage, probably.' Also with text and/or string keys the whole duplicate false/true loop logic could be simplified drastically (I think) to a containsKey / doesn't contain key if/else

src/main/java/dev/emi/trinkets/api/event/TrinketEquipCallback.java
fzzyhmstrs: As I thought, the original Drop callback had the exact indentation I used. without an exact guide for what indentation you want, I went with this:
image

But unsure where exactly you'd want the enclosing parentheses. Could also do this:
image

src/main/java/dev/emi/trinkets/api/event/TrinketEquipCallback.java
fzzyhmstrs: updated

src/main/java/dev/emi/trinkets/api/SlotAttributes.java
fzzyhmstrs: I will note that I'm not very satisfied with the method location, but there is no `impl` package as I've noted elsewhere, so I'm a bit at a loss where I'd put methods like this that a user realistically never needs. I really don't feel like adding an entire package from my end. But maybe if combined with the other methods in `Trinket`, it would be a useful refactor to make.

fzzyhmstrs:
Four comments I've left still unresolved; otherwise should be buttoned up now pending another look. One I have a bit of feedback I should be able to resolve everything open easily.

/////

As a side note to the PR, I find some of the documentation of API particulars to be less than ideal. `SlotAttributes#addSlotModifier` for example. I did not understand the implications of the method from the documentation:
image

`TrinketDropCallback#drop` (particularly what each choice of the enum does) and `TrinketApi#registerTrinketPredicate` (what behavior does TriState.DEFAULT provide? when should I use it over FALSE?) are some others that come to mind.

I bring it up because I would be happy to add some in the PR, but it's your documentation for your implementation details, so I am hesitant.

fzzyhmstrs:
My current thought re: the methods we aren't happy with is to add a `TrinketModifiers` class to the top-level package with the two `trinketModifiers` static methods and the `toSlotReferencedModifier` from `SlotAttributes`. 

Probably rename the `trinketModifiers` methods to something simpler, like `TrinketModifiers#get(args)`

If you are amenable to the idea I'll whip it up

src/main/java/dev/emi/trinkets/api/SlotAttributes.java
emilyploszaj: Doesn't nbt allow you to specify identifier per slot attribute? That is at least how it worked with uuids. It is sometimes desirable to have an accessory effect you cannot layer. Top level class `TrinketAttributes` with a handful of the static things we mentioned would be reasonable.

src/main/java/dev/emi/trinkets/api/SlotReference.java
emilyploszaj: Yeah I think `getId` is fine for now. I'll consider a further pass when I'm ready to merge that I can do on my own.

src/main/java/dev/emi/trinkets/api/event/TrinketEquipCallback.java
emilyploszaj: Yeah this is just style inconsistency in the codebase. Original is fine if it's matching some other similar class, apologies for the hard to resolve ask.

emilyploszaj:
The API/impl split of trinkets (and the general pattern for all of my projects) is everything in `dev.emi.trinkets.api` is API, everything else is impl. Re the general critiques about code style: a lot of style against my general preferences has leaked in over time, I'm only concerned with the stuff you're adding, anything else is my responsibility to go clean up some time. I think a top level `TrinketsModifiers` class with all the static helper stuff you've designed is reasonable to take it out of API space. Thanks for sticking with me in the review process.

src/main/java/dev/emi/trinkets/api/SlotAttributes.java
fzzyhmstrs: Yes, you can define slots, but if you *don't* define a slot, the previous implementation also locks you to one stack of the effect max, which seems unintentional.

emilyploszaj: Yes, but if you make this behavior the default you *cannot* make an effect with some uuid/id that is constant for all accessories, say, an accessory you can equip anywhere and gives you +10hp, but does not stack.

src/main/java/dev/emi/trinkets/api/event/TrinketEquipCallback.java
fzzyhmstrs: kept original

src/main/java/dev/emi/trinkets/api/Trinket.java
fzzyhmstrs: moved per conversation above/below

fzzyhmstrs:
Everything should be buttoned up in the latest commit.

src/main/java/dev/emi/trinkets/api/SlotAttributes.java
fzzyhmstrs: Unsure how to resolve it. One or the other functionality is lost entirely without further capabilities of the Component.

fzzyhmstrs: With the possibility of multiple slots per slot type, keeping the original "one-per-id" behavior seems less ideal. In the test mod for example, the "hat" adds 6 ring slots. So even if you define the ring slot in attribute NBT, you can only get one modifier despite the 7 identical slots; without some differentiation technique. 

In-code modifiers don't face this problem, they are given a unique id. based on slot reference id, which includes offset.

If only one slot was available per type, I could concede to vanilla behavior, which is one-per-id. But the behavior in-slot with multiple like-slots was one of my drivers to make the change in the first place, especially considering that it doesn't align to standard in-code behavior.

fzzyhmstrs:
re; the unresolved conversation above, I tested adding a `unique` field to the `TrinketsAttributeModifiersComponent`. By default this flag is `true`, which uses vanilla behavior, and leaves the id untouched. The user can set `unique:false` to have a modifier that works in multiples in either the same slot or various slots.

Only issue is, this would be undocumented behavior compared to vanilla attributes. 

image

image

emilyploszaj:
I think I'm going to maintain here that the attributes should have parity with vanilla, and the uniqueness strategies they have. If they've changed it from how they do uuids then that can be how trinkets does it.

src/main/java/dev/emi/trinkets/api/event/TrinketEquipCallback.java
```diff
@@ -0,0 +1,26 @@
+package dev.emi.trinkets.api.event;
+
+import dev.emi.trinkets.api.SlotReference;
+import dev.emi.trinkets.api.Trinket;
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ItemStack;
+
+public interface TrinketEquipCallback {
```
emilyploszaj, line 10: This class is indented using spaces, please use tabs

src/main/java/dev/emi/trinkets/api/event/TrinketUnequipCallback.java
```diff
@@ -0,0 +1,26 @@
+package dev.emi.trinkets.api.event;
+
+import dev.emi.trinkets.api.SlotReference;
+import dev.emi.trinkets.api.Trinket;
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ItemStack;
+
+public interface TrinketUnequipCallback {
```
emilyploszaj, line 10: This class is indented using spaces, please use tabs

src/main/java/dev/emi/trinkets/TrinketModifiers.java
```diff
@@ -0,0 +1,43 @@
...
+import com.google.common.collect.Multimap;
+import dev.emi.trinkets.api.*;
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.attribute.EntityAttribute;
+import net.minecraft.entity.attribute.EntityAttributeModifier;
+import net.minecraft.item.ItemStack;
+import net.minecraft.registry.entry.RegistryEntry;
+import net.minecraft.util.Identifier;
+
+public class TrinketModifiers {
```
emilyploszaj, line 12: This class is indented using spaces, please use tabs

src/main/java/dev/emi/trinkets/TrinketModifiers.java
```diff
@@ -0,0 +1,43 @@
+package dev.emi.trinkets;
+
+import com.google.common.collect.Multimap;
+import dev.emi.trinkets.api.*;
```
emilyploszaj, line 4: No star imports please

emilyploszaj (changes requested):
Very nearly mergeable, thanks for the changes

src/main/java/dev/emi/trinkets/TrinketModifiers.java
fzzyhmstrs: ok

src/main/java/dev/emi/trinkets/TrinketModifiers.java
fzzyhmstrs: ok

src/main/java/dev/emi/trinkets/api/event/TrinketUnequipCallback.java
fzzyhmstrs: ok

src/main/java/dev/emi/trinkets/api/event/TrinketEquipCallback.java
fzzyhmstrs: ok

fzzyhmstrs:
Reverted attribute behavior to vanilla style, removing the helper method.

Figured out some new things about intellij to make it stop doing star imports and space-tabs. Neat.

Hopefully all buttoned up now.

emilyploszaj (approved):

PR #369, opened 2025-02-25
Clear TrinketScreenManager on inventory close.

Ampflower:
Prevents potentially leaking worlds or entities.

This also affects 1.21.1 and older; but there's not a branch I can PR to for 1.21.

The alternate route you could take here is using WeakReference<TrinketScreen>, or perhaps combining the two.

'How does this fix a memleak?'

When combined with owo, it stores the player in the inventory object, which causes it to end up retaining an entire world via the player stored in a long since otherwise discarded inventory screen when it is stored by Trinkets. By having the inventory clear itself from the manager, it prevents a memleak by simply not allowing it to be retained for longer than necessary.

### DIFF
--- a/src/main/java/dev/emi/trinkets/TrinketScreenManager.java
+++ b/src/main/java/dev/emi/trinkets/TrinketScreenManager.java
@@ -1,5 +1,6 @@
 package dev.emi.trinkets;
 
+import java.lang.ref.WeakReference;
 import java.util.List;
 
 import com.mojang.blaze3d.systems.RenderSystem;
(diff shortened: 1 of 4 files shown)


Ampflower:
Oh, apparently this doesn't work quite as expected in production.

Ampflower:
Opting to push the WeakReference version, as that's been reliably working here.

The only other option I can think of is having the MinecraftClient call if the screen is set to null.

emilyploszaj (approved):
Seems like an owo problem but... sure

PR #390, opened 2025-06-20
Update to 1.21.7 (compatible with 1.21.6)

Patbox:
Biggest change is how data is saved/loaded. It now uses intermediary object, as it needed to go through codecs now. Reason why is that MC no longer uses nbt directly for saving and loading, but instead moving to Write/Read View wrapper classes. This also applies to synchronized data, through it still is baked by slightly modified save.

Bit smaller but still notable, there were some changes / moving of how the gui is rendering too (as Mojang also refactored that too), eliminating usage of z translations within the user code. Everything should look the same as in 1.21.5 through.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,17 +1,17 @@
 # Done to increase the memory available to gradle.
 org.gradle.jvmargs=-Xmx2G
 
-minecraft_version=1.21.5
-yarn_mappings=1.21.5+build.1
-loader_version=0.16.10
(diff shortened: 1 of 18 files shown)


vico93:
@Ryhon0 could you give a look on this PR for Trinkets-Canary (if Emily didn't publish herself)?

Patbox:
Updated dependencies and mc version to 1.21.7. It should be still 100% compatible with 1.21.6 through

PR #410, opened 2026-04-12
1.20.1 Slot Modification and GUI fixes

Aquaeyes:
This PR was initially intended to be relatively minor, but we kept bumping against related issues to fix. If splitting this into several PRs would be preferable, we can attempt to do so.

First, the methods for adding and removing attribute modifiers of a stack that has been unequipped was moved out of the LivingEntityMixin; This allows it to be called by the TrinketComponent refresh for inventory size changes and the Inventory Update function for when a slot modification causes *another* slot-modifying trinket to drop, additionally calling the relevant onUnequip method. Previously, if a trinket would drop from slot modification, then it's attributes would remain.

_As such, you could equip a trinket in the ring/0 slot that gives +1 ring slot, equip another in ring/1, and a third in ring/2. Removing ring/0 will still leave three ring slots, so no items would drop, down from the 4 you'd have with all three equipped. Removing ring/1 will reduce the slot count to 2, meaning that ring/2 will be dropped upon Inventory Update. However, without this change, ring/2's +1 ring slot removal would not process, leaving the player with 2 ring slots. Now, if the bottom 'block of the Jenga Tower' is removed, all the above 'blocks' will come crashing down too, instead of staying in place mid-air._

The two private methods were also merged, so as to allow a self-supporting trinket to be swapped in-place.

An additional check was added to the equipment change checks to not attempt to tick an item that has been removed from a slot which it supports - as said slot will no longer exist by the time that section of the method is reached, causing a crash.

Whilst not directly related to slot modification not being processed, MC-272769 was mitigated by adding a check upon unequipping an item with a static, non-slot-modified UUID intended to be exclusive. This checks that no other trinkets equipped have that attribute modifier before removing it, preventing the case where, if two trinkets are equipped with an attribute with the same UUID so as to not layer the effect, and one of them is removed, then the attribute is removed even though another trinket exists which provides that same attribute modifier UUID. ( Non-layerable UUIDs are presumed to be intended from here: https://github.com/emilyploszaj/trinkets/pull/320#discussion_r1649467685 )

One of the main issues with recursively modifying slot attributes, and therefore modifying the inventory, is that slot modification will mark an update, and any access to the inventory will run an update if marked, resulting in, if operation A would cause operation B to be run, for operation B to run *before* operation A's modifications would be saved as the inventory, meaning B's changes are overwritten when A resolves, and B doesn't see A's changes. As such, if update() is called mid-update, the update is suppressed until the outer update has resolved, at which point, it's triggered. This essentially turns a nesting-doll of resolves that never benefit from each other into tail-end recursion, resolving in a series in the order by which the operations are called. This ensures all modifications are kept, and taken into account, instead of being dropped or not detected.

Whilst also not directly related to slot count modification, if a trinkets slot is in a group with enough slots to extend past the natural trinket bounds, then it's considered as *outside* the bounds - therefore, left clicks will cause one item of that stack to be dropped, instead of the slot being naturally interactable. A trinket slot is now considered to be within trinket bounds, even if it extends past the static slots, to resolve this issue. If the number of slots in a slot type currently hovered over changes, the bounds are cleared, causing them to be recalculated once to check to see if the area has changed, meaning that the slot type doesn't need to be re-hovered in order to access slots previously outside the bounds, and that slots that no longer exist won't occlude slots beneath. Finally, the recipe book previously had uninteractable areas where trinket slots lay if the screen was in narrow mode - this has also been resolved.

In order to more easily achieve the above, a modified version @Ampflower's commit to the Main branch was merged into this one, which should also resolve the memory leaks, using the initial idea for the implementation due to other mods relying on the field type in this version.

Recursive updates were changed to use the old stacks present, as the method in LivingEntityMixin does, through the interface LivingEntityTrinketComponent.StackHistory. A list of dropped stacks is kept in the update method of a LivingEntityTrinketComponent whilst it processes items dropped, so as to modify the new inventory when processing Attribute Modifiers and onUnequips, and removed slots will also attempt to drop their items. Due to the potential for odd behaviour in onUnequip implementations acting upon a slotType that no longer exists, that section is in a try-catch block.

I am certain that large sections of this do not adhere to the style of the mod, and will gladly refactor to improve code quality if needed. There are quite likely some crashes and strange behaviours that have been overlooked, but after some testing and finding none, hopefully they are far less consequential or harder to trigger than the ones that this PR fixes? I have attempted to maintain compatibility, additionally, and have attempted to avoid changes to method signatures that may be accessed by other mods. The main method changed, the private methods in LivingEntityMixin, *shouldn't* be directly referenced by other mods, and, additionally, is not in the API package, so our assumption is that those were safe to change. If we need to elaborate on any points here, we would be more than happy to do so.

~~**NOTE**: When building locally, lambda identifiers are swapped from 3.7.2 - e.g. lambda$render$2 and lambda$render$0 are swapped in TrinketFeatureRenderer. This breaks PeculiarPieces and TrinkHide, for example.~~ Resolved via clean build.

### DIFF
--- a/src/main/java/dev/emi/trinkets/TrinketScreen.java
+++ b/src/main/java/dev/emi/trinkets/TrinketScreen.java
@@ -20,6 +20,10 @@ public interface TrinketScreen {
 		return false;
 	}
 
+	public default boolean trinkets$isNarrow() {
+		return false;
+	}
+
 	public default void trinkets$updateTrinketSlots() {
 	}
 }
--- a/src/main/java/dev/emi/trinkets/TrinketScreenManager.java
+++ b/src/main/java/dev/emi/trinkets/TrinketScreenManager.java
@@ -25,7 +25,6 @@ public class TrinketScreenManager {
 	public static SlotGroup quickMoveGroup = null;
 
 	public static void init(TrinketScreen screen) {
-		System.out.println(screen);
 		currentScreen = screen;
 		group = null;
 		currentBounds = new Rect2i(0, 0, 0, 0);
--- a/src/main/java/dev/emi/trinkets/TrinketScreenManager.java
+++ b/src/main/java/dev/emi/trinkets/TrinketScreenManager.java
@@ -316,6 +316,10 @@ public class TrinketScreenManager {
 
 	public static boolean isClickInsideTrinketBounds(double mouseX, double mouseY) {
 		TrinketPlayerScreenHandler handler = currentScreen.trinkets$getHandler();
+		if (currentScreen.trinkets$getFocusedSlot() instanceof TrinketSlot) {
+			return true;
+		}
+
 		int x = currentScreen.trinkets$getX();
 		int y = currentScreen.trinkets$getY();
 		int mx = (int) (Math.round(mouseX) - x);
--- a/src/main/java/dev/emi/trinkets/TrinketScreenManager.java
+++ b/src/main/java/dev/emi/trinkets/TrinketScreenManager.java
@@ -123,9 +123,11 @@ public class TrinketScreenManager {
 					continue;
 				}
 				if (r.contains(Math.round(mouseX) - x, Math.round(mouseY) - y)) {
-					TrinketsClient.activeGroup = g;
-					TrinketsClient.quickMoveGroup = null;
-					break;
+					if (!(currentScreen.trinkets$isNarrow() && currentScreen.trinkets$isRecipeBookOpen())) {
+						TrinketsClient.activeGroup = g;
+						TrinketsClient.quickMoveGroup = null;
+						break;
+					}
 				}
 			}
 		}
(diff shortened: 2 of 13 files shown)


Aquaeyes:
Latest 3 commits add a callback for SlotCount modification of a TrinketComponent's inventories via EAM - internally, this just contains the contents of `TrinketComponent::getTrackingUpdates` - but clearing the accumulated inventories has been moved out of the client - otherwise, the list of inventories would slowly accumulate clientside, never being cleared, especially if the TrinketComponent was updated, as the new inventories would exist alongside the previous ones of the same SlotType.

Additionally, Entities update their TrinketComponents on datapack reload, so removing a slot will cause the item to drop now, instead of on the next time the entity loads. This does iterate all entities in all dimensions, however, but only on datapack reload. TrinketScreens / TrinketPlayerScreenHandlers that a player has open are also updated if they do not belong to a player on datapack reload.

Of these, I'm the least sure about about the Screen/ScreenHandler update on datapack reload. Whilst it shouldn't do any harm, and extra ScreenHandlers *should* be updated on datapack reload, there are some situations in which it can update the screen itself twice unnecessarily. There may be a more elegant solution?

